Issue #45, opened 2023-04-13
High CPU usage on Linux


tkolo:
I've been using this SDK under Arch Linux and encountered an issue similar to #13 and #37.
Upon closer inspection the problem seems to lie in `LinuxHidDeviceListener.CheckForUpdates`, here: https://github.com/Yubico/Yubico.NET.SDK/blob/1.7.0/Yubico.Core/src/Yubico/Core/Devices/Hid/LinuxHidDeviceListener.cs#L157
It seems that the code expects the call to `udev_monitor_receive_device` to block but, at least on my system, it doesn't, instead returning NULL pointer immediately. 
I'm by no means familiar with udev and systemd's documentation leaves a lot to be desired so I can't tell whenever this is a bug in udev/systemd or this SDK.

I'm using Arch Linux with systemd 253
```
$ systemctl --version
systemd 253 (253.3-3-arch)
+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT -QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified
```

GregDomzalski:
Apologies for the delay. The fix for this issue has now been released in our 1.8.0 release: https://github.com/Yubico/Yubico.NET.SDK/releases/tag/1.8.0

I'm resolving this issue now that the fix should be in our public release.

PR #46, opened 2023-04-14
Added method to remove of O_NONBLOCK flag from the monitor's udev socket descriptor. 

jkolo:
Resolve issue #45


GregDomzalski:
Thanks for submitting the PR. The team will look at this in the next few weeks. I really appreciate the contribution!

GregDomzalski (approved):

jkolo:
I note that a signed commit is required. I added the signature and pushed the changes. I look forward to seeing the release of the patched version.

GregDomzalski:
Much appreciated!

Our regular release cadence is quarterly - so we just missed that window. We do however usually issue some patch releases in the interim once we have enough fixes that we want to take out.

I would anticipate a few weeks until we get a signed release out. If you need this sooner, please let me know and I can see what our options are.

PR #58, opened 2023-09-08
Update MSBuild DLL import conditions

ThatOneCamel:
This fixes an issue that causes the wrong Yubio.NativeShims.dll to be used. If a user uses the 'AnyCPU' platform target, this issue could potentially result in users mixing 32-bit and 64-bit dlls and leave users with errors such as 'BadImageFormatException' whenever code that utilizes the Yubico SDK was called.

AnyCPU shouldn't default to x86 unless the architecture of the host machine performing a build is also x86.
AnyCPU can also utilize x64 architecture, and is supposed to only fallback to x86 if x64 isn't possible.
Defaulting to either option can potentially result in bitness mismatches

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/output#platformtarget

### DIFF
--- a/Yubico.NativeShims/msbuild/Yubico.NativeShims.targets
+++ b/Yubico.NativeShims/msbuild/Yubico.NativeShims.targets
@@ -3,14 +3,14 @@
     <!-- .NET Framework versions don't seem to include the native runtime files that
         are needed for this package. This makes sure they are included in downlevel projects.
         Since .NET Framework is Windows only, we only need to worry about that platform. -->
-    <ItemGroup Condition="'$(Platform)' == 'x86' OR '$(Platform)' == 'AnyCPU'">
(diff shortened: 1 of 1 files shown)


GregDomzalski:
Thanks for the PR - apologies for the delay in looking at this. Most of the team has been out for a work conference. We will take a look when we get back. This is something that we will likely take. Thank you for your contribution.

GregDomzalski (approved):

GregDomzalski:
@ThatOneCamel Is there any way you can sign the commit for us?

If not, it's a small enough change that I'd be comfortable signing it using my key.

Please LMK asap - as I want to try to get it into today's release.

ThatOneCamel:
@GregDomzalski The commit has been signed

GregDomzalski:
Thanks!

PR #75, opened 2024-04-12
Set up dotnet in dependent workflows

DennisDyallo:
Sometimes the dotnet format workflow would fail unexpectedly when the runner couldnt locate the dotnet executable. This PR attempts to address that matter.


DennisDyallo (commented):
These are just workflow adjustments, so I will go ahead merge them.

PR #89, opened 2024-05-15
Make YubiKeyDeviceListener resettable

jamiehankins:
# Description

Currently, YubiKeyDeviceListener only exposes a singleton instance with no way for a consumer to reset it once it is started. The constructor is private, so the consumer has no option to use a private instance.

This adds the ability for the consumer to reset the lazy-initialized singleton so that the listen thread doesn't stick around.

Also, even though this class implemented IDisposable, most things that need to be in the Dispose method weren't there, including a way to stop the thread.

The listen thread is now replaced by an asynchronous task that doesn't consume a system thread while it's not working.

Fixes # YESDK-1313

## Type of change

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

**Test configuration**:
* Firmware version: 5.4.3
* Yubikey model:

# Checklist:

- [ ] My code follows the [style guidelines](https://raw.githubusercontent.com/Yubico/Yubico.NET.SDK/043119ad1d19e0e6e66556c970a81d0c1aba36c8/CONTRIBUTING.md) of this project 

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
@@ -17,6 +17,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
 using Yubico.Core.Devices;
 using Yubico.Core.Devices.Hid;
 using Yubico.Core.Devices.SmartCard;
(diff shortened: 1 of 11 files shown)


DennisDyallo (commented):
It looks good to me. Because of limited, I don't feel confident in testing this PR myself. Perhaps @GregDomzalski will have some time to take a look ?

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
```diff
@@ -44,83 +45,84 @@ public class YubiKeyDeviceListener : IDisposable
         /// <summary>
         /// An instance of a <see cref="YubiKeyDeviceListener"/>.
         /// </summary>
-        public static YubiKeyDeviceListener Instance => _lazyInstance.Value;
+        public static YubiKeyDeviceListener Instance => _lazyInstance ??= new YubiKeyDeviceListener();
 
-        private static readonly Lazy<YubiKeyDeviceListener> _lazyInstance =
-            new Lazy<YubiKeyDeviceListener>(() => new YubiKeyDeviceListener());
+        /// <summary>
+        /// Disposes and closes the singleton instance of <see cref="YubiKeyDeviceListener"/>.
```
GregDomzalski, line 51: Summary: Instructs the SDK to stop listening for YubiKey events, effectively causing the SDK to go "dormant". Accessing `Instance` or calling `YubiKeyDevice.FindAll()` will restart the listener.

Remarks: Enumerating YubiKeys is actually done via a cache. As such, this cache must be maintained and kept up-to-date. This is done by starting several listening threads that run in the background. These listen for the relevant OS device arrival and removal events. Normally, these background listeners will run starting with the first enumeration call to the SDK and remain active until the process shuts down (i.e. a lazy singleton). But there are cases where you may not want the overhead of these listeners running all the time. While they do their best to not eat CPU, they can sometimes generate log noise, exceptions, etc. This method allows you to stop these background listeners and reclaim as much resources as possible from the SDK. This will not invalidate any existing `IYubiKeyDevice` instances, however you will not receive any additional events regarding that device. Any subsequent calls to `YubiKeyDevice.FindAll`, `YubiKeyDevice.FindByTransport`, or `YubiKeyDeviceListener.Instance` will restart the listening threads.


Suggested documentation updates describing the impact to the caller rather than what's it's doing under the covers.

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
```diff
@@ -44,83 +45,84 @@ public class YubiKeyDeviceListener : IDisposable
...
             while (_isListening)
             {
-                _ = updateEvent.WaitOne();
-                Thread.Sleep(200); // I really dislike sleeps, but here, it does seem like a good idea to give the
-                                   // system some time to quiet down in terms of PnP activity.
-                _ = updateEvent.Reset();
-                Update();
+                try
+                {
+                    await _semaphore.WaitAsync(CancellationToken).ConfigureAwait(false);
```
GregDomzalski, line 109: We dropped a log message here. Knowing what started the update process is very useful to know.

GregDomzalski (commented):
I think we need to do the same thing for the HidDeviceListener and SmartCardDeviceListener classes. Otherwise we're not _really_ stopping anything or freeing up threads/resources.

DennisDyallo (changes requested):
Hi @jamiehankins ! 
Just writing to let you know that I'm still aware of this case but that it's beeing put on the back burner in favor of higher priority cases. If you have the time, please feel free to look through @GregDomzalski comments to implement this fix in a more comprehensive way. I promise to review any upcoming amendments as quickly as possible.
Have a good one!

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
```diff
@@ -349,23 +351,46 @@ private static IEnumerable<IDevice> GetSmartCardDevices()
...
-            if (!_disposedValue)
+            if (!_isDisposed)
             {
                 if (disposing)
                 {
-                    RwLock.Dispose();
+                    // Signal the listen thread that it's time to end.
+                    _tokenSource.Cancel();
+
+                    // Shut down the listener handlers.
```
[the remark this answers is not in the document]
jamiehankins: @GregDomzalski I'm not sure I see what you mean. We're not just stopping the thread, we're disposing the `YubiKeyDeviceListener` instance. Its `Dispose` method unsubscribes from the HID and smartcard listeners, and if they implement `IDisposable`, disposes of them. The references to them die when GC collects the YubiKeyDeviceListener instance. The finalizers all call `StopListening` in one way or the other.

I suppose one step further we could go is to make all of the listeners IDisposable (the HID ones are not currently) so that we can hasten their demise, and leave the finalizer so that legacy code doesn't get broken. I don't really think this is necessary, though.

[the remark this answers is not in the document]
GregDomzalski: Hmm yeah. I see what you mean. I must have looked at the HID listeners when originally making this comment. It seems very strange to me that we chose to implement a finalizer only and not `IDisposable`. That doesn't really seem very kosher. But it is what it is, and I agree - I believe your code is doing the right thing based on what's present elsewhere. Feel free to resolve this thread.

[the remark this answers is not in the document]
jamiehankins: @DennisDyallo you good with merging to develop?

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
jamiehankins: Added logging.

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDeviceListener.cs
jamiehankins: I used this as a basis, but removed references to "threads", since this listener doesn't use threads anymore, and the HID listener only subscribes to whatever the OS does.

DennisDyallo (approved):

PR #93, opened 2024-05-21
Fast USB interface reclaim

GregDomzalski:
# Description

The reclaim timeout was introduced in the YubiKey NEO. Because the NEO used a separate chip for USB communication from the secure element, there was a delay that was needed for stability reasons whenever the host (computer) wanted to switch USB interfaces. This interface "reclaim" timeout was a whopping 3 seconds - meaning if you wanted to switch from talking to OTP to FIDO, you needed to wait 3 seconds prior to issuing the first FIDO command.

Even though it was no longer needed in the YubiKey 4/5 hardware platform, this reclaim timeout behavior was ported over. After some discussions with the firmware team, they agreed that this timeout was no longer needed and could be removed. This "fast reclaim" was merged into FW 5.6.0 and is available in all subsequent versions.

This PR adds a feature flag and check and will default to no longer waiting for the 3 seconds if we detect a key with this ability. The behavior can be overridden by an app compat switch in case anyone was depending on this old timing.

Although we were already taking shortcuts during enumeration to avoid talking to the same YubiKey more than once, this will speed up the degenerate cases where those shortcuts are not possible.

This also drastically improves performance on configuration and provisioning style applications that need to switch between OTP, SmartCard, and FIDO interfaces quickly.

## Type of change

- [x] This change requires a documentation update

# How has this been tested?

Existing integration test was updated to always force the old behavior. The test was run using both a 5.4.3 YubiKey and a 5.7.x YubiKey. The reclaim timeout behavior remained. Since we don't yet have a `StandardTestDevice` filter for very new keys like `5.7` I did not add any additional test cases. However I did modify the test locally to verify that fast reclaim worked by default on 5.7, that slow reclaim worked by default on 5.4.3, that slow reclaim worked when overridden for both 5.7 and 5.4.3 when the app compat flag is set.

**Test configuration**:
* Firmware version: 5.4.3, 5.7.1
* Yubikey model: 5C NFC

# Checklist:

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/FirmwareVersion.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/FirmwareVersion.cs
@@ -20,8 +20,8 @@ namespace Yubico.YubiKey
     public class FirmwareVersion : IComparable<FirmwareVersion>, IComparable, IEquatable<FirmwareVersion>
     {
         #region Frequently Used Versions
-        // Note that these are for internal use. Later, we want to have something
-        // that allows users to query for specific capabilities instead of versions.
+        // Note that these are for internal use. It's expected that SDK users should call `.HasFeature` on the
+        // YubiKey device to check for features instead of FW versions.
         internal static readonly FirmwareVersion All = new FirmwareVersion(1, 0, 0);
         internal static readonly FirmwareVersion V2_0_0 = new FirmwareVersion(2, 0, 0);
         internal static readonly FirmwareVersion V2_1_0 = new FirmwareVersion(2, 1, 0);
@@ -40,6 +40,7 @@ namespace Yubico.YubiKey
         internal static readonly FirmwareVersion V5_3_0 = new FirmwareVersion(5, 3, 0);
         internal static readonly FirmwareVersion V5_4_2 = new FirmwareVersion(5, 4, 2);
         internal static readonly FirmwareVersion V5_4_3 = new FirmwareVersion(5, 4, 3);
+        internal static readonly FirmwareVersion V5_6_0 = new FirmwareVersion(5, 6, 0);
         #endregion
 
         public byte Major { get; set; }
--- a/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyCompatSwitches.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyCompatSwitches.cs
@@ -1,0 +1,29 @@
+// Copyright 2024 Yubico AB
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// You may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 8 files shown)


Yubico.YubiKey/src/Yubico/YubiKey/FirmwareVersion.cs
```diff
@@ -40,6 +40,7 @@ public class FirmwareVersion : IComparable<FirmwareVersion>, IComparable, IEquat
         internal static readonly FirmwareVersion V5_3_0 = new FirmwareVersion(5, 3, 0);
         internal static readonly FirmwareVersion V5_4_2 = new FirmwareVersion(5, 4, 2);
         internal static readonly FirmwareVersion V5_4_3 = new FirmwareVersion(5, 4, 3);
+        internal static readonly FirmwareVersion V5_6_0 = new FirmwareVersion(5, 6, 0);
```
DennisDyallo, line 43: Should we add 5.7.0 to this list right away?

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyCompatSwitches.cs
DennisDyallo, line 1: I'm thinking maybe we should document these switches somewhere...

https://github.com/Yubico/Yubico.NET.SDK/blob/develop/Yubico.Core/src/Yubico/Core/CoreCompatSwitches.cs

DennisDyallo: Perhaps here somewhere?

image

DennisDyallo (approved):
Looks good to me, tests are passing as well. If you could suggest on a location to document the switches and add the information there Id say its a thumbs up and you can merge when ready

AdamVe (approved):
LGTM

Yubico.YubiKey/src/Yubico/YubiKey/FirmwareVersion.cs
GregDomzalski: Not in this PR 😄 I would say add it when you need it.

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyCompatSwitches.cs
GregDomzalski: Sounds good. Will make drop a doc there.

Yubico.YubiKey/src/Yubico/YubiKey/FirmwareVersion.cs
DennisDyallo: Ok, I just did :)

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyCompatSwitches.cs
DennisDyallo: Awesome :)

PR #110, opened 2024-06-25
Add device reset command

AdamVe:
# Description

Adds command for device reset

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

**Test configuration**:
* Firmware version:
* Yubikey model:

# Checklist:

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/IYubiKeyDevice.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/IYubiKeyDevice.cs
@@ -693,5 +693,15 @@ namespace Yubico.YubiKey
         /// </remarks>
         /// <param name="value">The touch threshold to apply to the YubiKey. Must be a value greater than `6`.</param>
         void SetTemporaryTouchThreshold(int value);
+
+        /// <summary>
+        /// Perform device wide reset.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Only available on YubiKey Bio Multi-protocol.
+        /// </para>
+        /// </remarks>
+        void DeviceReset();
     }
 }
--- a/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDevice.Instance.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDevice.Instance.cs
@@ -741,6 +741,45 @@ namespace Yubico.YubiKey
             }
         }
 
+        /// <inheritdoc />
+        /// <exception cref="NotSupportedException">
+        /// The YubiKey does not support this feature.
+        /// </exception>
+        /// <exception cref="InvalidOperationException">
+        /// The YubiKey encountered an error and could not set the setting.
+        /// </exception>
+        public void DeviceReset()
+        {
+            if (!this.HasFeature(YubiKeyFeature.DeviceReset))
+            {
+                throw new NotSupportedException(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        ExceptionMessages.NotSupportedByYubiKeyVersion));
+            }
+            IYubiKeyConnection? connection = null;
+            try
+            {
+                if (TryConnect(YubiKeyApplication.Management, out connection))
+                {
+                    var resetCommand = new MgmtCmd.DeviceResetCommand();
+                    IYubiKeyResponse response = connection.SendCommand(resetCommand);
+                    if (response.Status != ResponseStatus.Success)
+                    {
+                        throw new InvalidOperationException(response.StatusMessage);
+                    }
+                }
+                else
+                {
+                    throw new NotSupportedException(ExceptionMessages.NoInterfaceAvailable);
+                }
+            }
+            finally
+            {
+                connection?.Dispose();
+            }
+        }
+
         private IYubiKeyResponse SendConfiguration(MgmtCmd.SetDeviceInfoBaseCommand baseCommand)
         {
             IYubiKeyConnection? connection = null;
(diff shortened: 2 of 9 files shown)


DennisDyallo:
Is this ready for review @AdamVe ?

Yubico.YubiKey/src/Yubico/YubiKey/IYubiKeyDevice.cs
```diff
@@ -693,5 +693,15 @@ public interface IYubiKeyDevice : IYubiKeyDeviceInfo, IEquatable<IYubiKeyDevice>
...
+
+        /// <summary>
+        /// Perform device wide reset.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Only available on YubiKey Bio Multi-protocol.
+        /// </para>
+        /// </remarks>
+        void DeviceReset();
```
DennisDyallo, line 705:
```suggestion
        void ResetDevice();
```

We like to name things so that they make sense semantically, if we can use imperative form here it adds to the semantics

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyDevice.Instance.cs
```diff
@@ -741,6 +741,45 @@ public void SetTemporaryTouchThreshold(int value)
...
+                        CultureInfo.CurrentCulture,
+                        ExceptionMessages.NotSupportedByYubiKeyVersion));
+            }
+            IYubiKeyConnection? connection = null;
+            try
+            {
+                if (TryConnect(YubiKeyApplication.Management, out connection))
+                {
+                    var resetCommand = new MgmtCmd.DeviceResetCommand();
+                    IYubiKeyResponse response = connection.SendCommand(resetCommand);
```
DennisDyallo, line 766:
```suggestion
                    var command = new MgmtCmd.DeviceResetCommand();
                    IYubiKeyResponse response = connection.SendCommand(command);
```

Yubico.YubiKey/src/Yubico/YubiKey/IYubiKeyDevice.cs
DennisDyallo: Coming back to this, I agree that it makes sense to keep the same names as the other SDKs

DennisDyallo:
LGTM!

DennisDyallo (approved):
LGTM

Yubico.YubiKey/src/Yubico/YubiKey/YubiKeyFeature.cs
```diff
@@ -74,6 +74,11 @@ public enum YubiKeyFeature
         /// </summary>
         TemporaryTouchThreshold,
 
+        /// <summary>
+        ///     The YubiKey supports proprietary reset.
```
equijano21, line 78: @AdamVe what does "proprietary" mean here exactly?

AdamVe: Great catch! The _propietary_ in here is referring to a function of the YubiKey which is only available on the YubiKey (and is not part of the PIV standard). The reasoning is:

When both applications (FIDO and PIV) are enabled AND configured with credentials, a _Yubico proprietary factory reset_ will be the only way to reset both FIDO and PIV, simultaneously.

Maybe instead of "proprietary" it is better to use "device wide factory" because it will reset all the applications (FIDO and PIV in this context) on the device:

```
The YubiKey supports device wide factory reset.
```

equijano21: Ah ok, interesting. So does this mean that a reset that only affects the PIV application (Yubico.YubiKey.Piv.PivSession.ResetApplication) adheres to the PIV standard? 

And if this reset only affects the FIDO and PIV applications, shouldn't we specify that? When I read "device-wide", I would assume that affects all YubiKey applications (Yubico OTP, etc).

AdamVe: Oh, I have forgotten to mention that this instruction is _available only on YubiKey Bio Multi-Protocol Edition devices_. There are only FIDO and PIV application available on these devices. These application share the PIN and the fingerprint sensor (you can use the same pin/fingerprints for both FIDO and PIV authentication) which causes that you cannot reset the PIV or FIDO applications with the "PivSession.ResetApplication" (and vice versa for FIDO) because it would invalidate the other application as well. That is why the device wide reset was created - it will reset all the applications on the YubiKey (currently FIDO and PIV). 

Once you do some setup of FIDO (for example add passkeys) or PIV (for example import a certificate), the PivSession.ResetApplication (or the FIDO reset) will not be possible to call because of the reason I described in the previous paragraph. The only way how to reset such YubiKey is to use the device wide reset.

I think it should be added to the documentation that this is available only on YubiKey Bio MPE.

equijano21: Very cool, I didn't know that! 

I added some additional info about the DeviceReset in my [PR](https://github.com/Yubico/Yubico.NET.SDK/pull/126) in this commit: https://github.com/Yubico/Yubico.NET.SDK/pull/126/commits/3b234d5a76affc49865d815843098063d65221a0. Let me know if this covers it sufficiently.

AdamVe: Great! That is much better. I added a comment there to clarify a detail. Thank you :)

PR #132, opened 2024-07-30
Using PolySharp to allow use of C#13

DennisDyallo:
# Description

This update removes the need for the Yubico.DotnetPolyfills-project. This project was adding a couple of methods, such as the Range and the Index operator. Now, since adding Polysharp, this is no longer needed as Polysharp generates polyfills for newer C# language features and APIs at compile-time. This also allows us to use C# [language version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning#defaults) 13 instead of 7.3.

Examples of new syntax can be seen [here](https://github.com/Yubico/Yubico.NET.SDK/blob/dennisdyallo/advanced-syntax/Yubico.YubiKey/src/Yubico/YubiKey/CSharpLangFeaturesExamples.cs) in a stowed away branch in this repo.

> Note
It's true that bringing in a dependency into any library should not be taken in lightly. This suggestion has been with the team for some time, and while we haven't committed to it yet, we want to explore the possibility of making such a change.  We do believe measures can be taken to ensure stability and expected behavior. Rest assured, although the third party dependency in question appears to be well regarded, useful and stable, it won't be added without going through a due dilligence process.

To anyone else reading this, please feel free to add your comments or any alternatives to move the SDK forward in this regard.

Fixes: YESDK1316

## Type of change

# How has this been tested?
Unit tests pass and builds pass.

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas


iamcarbon:
This introduces an untrusted third party dependency into a security critical library. It would be better to selectively port the individual Polyfills into core, then introduce a new dependency (and extend the chain of trust).

DennisDyallo:
> This introduces an untrusted third party dependency into a security critical library. It would be better to selectively port the individual Polyfills into core, then introduce a new dependency (and extend the chain of trust).

Hi @iamcarbon and thanks for your raised concern. It's true that bringing in a dependency into any library should not be taken in lightly. This suggestion has been with the team for some time, and while we haven't committed to it yet, we want to explore the possibility of making such a change. Rest assured, although the third party dependency in question appears to be well regarded, useful and stable. It won't be added without going through a due dilligence process. 

To you and anyone else reading this, please feel free to add your comments or any alternatives to move the SDK forward in this regard.

Yubico.YubiKey/tests/integration/Yubico/YubiKey/Piv/Commands/AuthMgmtKeyCmdTests.cs
```diff
@@ -88,7 +88,7 @@ public void AuthKey_Aes_Succeeds()
                     0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57, 0x58
                 };
 
-                var initCmd = new InitializeAuthenticateManagementKeyCommand(true, PivAlgorithm.TripleDes);
+                var initCmd = new InitializeAuthenticateManagementKeyCommand(true, PivAlgorithm.TripleDes); // TODO this test only works with 5.4.3 and lower
```
DennisDyallo, line 91: Will handle this in test rework case

DennisDyallo, line 91: Will handle this in a test rework case

dainnilsson (approved):

PR #134, opened 2024-08-05
Add support for enforceCredProtectPolicy in AddCredProtectExtension

AaFortner:
[The CTAP spec](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-credProtect-extension) defines two input parameters for the client-side part of the credProtect extension, credentialProtectionPolicy and enforceCredentialProtectionPolicy. MakeCredentialParameters.AddCredProtectExtension already supports handling the credentialProtectionPolicy parameter, but it does not support the enforceCredentialProtectionPolicy parameter. This means that callers who want to properly support enforceCredentialProtectionPolicy would need to do their own checks against the keys' support levels and the requested policy so as not to return an error when enforceCredProtectPolicy is false, and in doing so would end up duplicating most of what's in AddCredProtectExtension already.

This change add support for silently ignoring credentialProtectionPolicy when the YK does not support the extension if enforceCredentialProtectionPolicy is false.

Fixes # https://yubico.atlassian.net/browse/PROD-2309

## Type of change

I deliberately added enforceCredentialProtectionPolicy as an optional parameter with a default of true so as not to make this a breaking API change. That said, it does make the parameter order a little weird. If anyone feels strongly about it, another option would be to create an explicit function overload with a 'better' parameter order and have the old API call the new one (passing true for enforceCredentialProtectionPolicy to maintain the current behavior).

# How has this been tested?
These changes were tested both with manual testing and new integration tests that run through all of the following combination of values.

Test cases 
YK FW 5.4.3
Policy=optional, Enforce=true
Policy=required, Enforce=true
Policy=optional, Enforce=false
Policy=required, Enforce=false
In all four combinations, programming was successful, and extension data was present in the response's AuthenticationData

YK FW 5.1.2
Policy=optional, Enforce=true*
Policy=required, Enforce=true - Exception thrown from AddCredProtectExtension (as expected)
Policy=optional, Enforce=false*
Policy=required, Enforce=false*
In all but case two, programming was successful, and extension data was NOT present in the response's AuthenticationData (as expected)

**Test configuration**:
* Firmware versions: 5.4.3 and 5.1.2
* Yubikey models: NFC A 

# Checklist:

- [n/a] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/Fido2/MakeCredentialParameters.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/Fido2/MakeCredentialParameters.cs
@@ -645,8 +645,8 @@ namespace Yubico.YubiKey.Fido2
         /// </exception>
         public void AddCredProtectExtension(
             CredProtectPolicy credProtectPolicy,
-            AuthenticatorInfo authenticatorInfo,
-            bool enforceCredProtectPolicy = true)
+            bool enforceCredProtectPolicy,
+            AuthenticatorInfo authenticatorInfo)
         {
             if (credProtectPolicy == CredProtectPolicy.None)
             {
@@ -671,6 +671,11 @@ namespace Yubico.YubiKey.Fido2
             // values are 1, 2, or 3, so the encoding is simply 0x01, 02,or 03.
             AddExtension(KeyCredProtect, new byte[] { (byte)credProtectPolicy });
         }
+
+        /// <inheritdoc cref="AddCredProtectExtension(CredProtectPolicy,bool,AuthenticatorInfo)"/>
+        public void AddCredProtectExtension(CredProtectPolicy credProtectPolicy,
+                                            AuthenticatorInfo authenticatorInfo) =>
+            AddCredProtectExtension(credProtectPolicy, true, authenticatorInfo);
 
         /// <summary>
         /// Add an entry to the list of options.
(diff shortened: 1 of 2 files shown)


DennisDyallo:
Hi @AaFortner and thanks for the PR! I agree that the ordering is a bit weird, but I'm OK with it being that way rather than introduce a breaking change.

I see you've tested properly manually which is great. Would you consider adding an integration test? `Yubico.YubiKey/tests/integration/Yubico/YubiKey/Fido2/MakeCredential...` looks to be a good candidate. 
This would ensure that the logic would survive refactoring :smile:

AaFortner:
Thanks, Dennis! 

Here's a quick summary of the latest updates:
- It sounded like you didn't have strong feelings about the parameter order weirdness, but I went ahead and added an explicit overload with a nicer parameter order anyway. 
- Added integrations tests. Thanks for pushing these! In writing them, I discovered that my manual testing was slightly incorrect and exceptions were being thrown more often than strictly necessary on older FW. 
- Per the previous, slightly loosened the cases for where an exception is thrown on older FW
- Fixed a bug in one of the other integrations tests where the pin wasn't being set after reset

DennisDyallo (approved):
Fantastic. Looks good to me!

Issue #151, opened 2024-09-26
[BUG] BadImageFormatException when using the library in a .NET 4.7 project, compiled on a 64bit PC with AnyCPU and executed on 32bit PC.


romerod:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When the library is used in a project compiled with AnyCPU which is compiled on a x64 bit system the applications fails with a BadImageFormatException on 32bit PC as the Yubico.NativeShims.dll in the build output folder is the x64 version. The same thing happens when setting the Prefer32Bit flag.

The Yubico.NativeShims.targets file in the Yubico.NativeShims nuget package copies the file over depending of the architecture of the build system.

### Expected Behavior

When a project is build with AnyCPU the build output should contain the x86 and the x64 bit of the dll.

And the dll to load should be detected on runtime. For example by using the LoadLibrary method of kernel32.dll. For the moment we worked around the issue with copying both dlls to the build output and then with a code similar to this:

```
if(Environment.Is64BitProcess)   
    NativeMethods.LoadLibrary("x64\Yubico.NativeShims.dll")
else
   NativeMethods.LoadLibrary("x86\Yubico.NativeShims.dll")
```

before calling any of the library methods.

### Steps To Reproduce

1. Create .NET 4.72 project with AnyCPU
2. Add Yubico.YubiKey nuget package reference
3. Call YubiKeyDevice.FindByTransport(Transport.All);
4. Build it on a x64 PC and run it on a x32bit system

Can also be reproduced by setting the `<Prefer32Bit>true</Prefer32Bit>` instead of running it on 32Bit system.

### Version

1.11.0

### Version

5.4.3

DennisDyallo:
Thanks for the report. The fix will be added to the next release @romerod

romerod:
Thank you, looking forward to test it when it released 🎉

DennisDyallo:
These should be fixed now @romerod  , please get back to me if you run into any issues

PR #154, opened 2024-10-21
Implement dynamic DLL resolution of NativeShims for .NET 4.7 x86/x64 architectures

DennisDyallo:
# Description

Quote: @romerod

> **Current Behavior**
> When the library is used in a project compiled with AnyCPU which is compiled on a x64 bit system the applications fails with a BadImageFormatException on 32bit PC as the Yubico.NativeShims.dll in the build output folder is the x64 version. The same thing happens when setting the Prefer32Bit flag.
> 
> The Yubico.NativeShims.targets file in the Yubico.NativeShims nuget package copies the file over depending of the architecture of the build system.
> 
> **Expected Behavior**
> When a project is build with AnyCPU the build output should contain the x86 and the x64 bit of the dll, and the dll to load should be detected on runtime. 

Fixes include:
- Implementing dynamic DLL resolution based on process architecture (x86/x64)

Fixes: #151 

## Type of change

- [ ] Refactor (non-breaking change which improves code quality or performance)

## How has this been tested?

I built a NET47 32 bit application according to the #151 issue and reproduced the bug. Loaded the SDK dlls into that application and proceeded testing and fixing. 

**Test configuration**:
* OS version: Windows 11
* Firmware version: 5.4.3
* Yubikey model[^1]: 5C

## Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
+++ b/Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
+// As long as we have the Libraries.Net47.cs class which holds the opposite preprocessor directive check,
+// this check is required - as having both at the same time is not possible.
+#if !NET47 
 
 namespace Yubico.PlatformInterop
 {
@@ -35,34 +37,6 @@ namespace Yubico.PlatformInterop
     /// </remarks>
     internal static partial class Libraries
     {
-#if NET47
-        /// <summary>
-        /// The filename of the native shims library for .NET Framework 4.7.
-        /// </summary>
-        /// <remarks>
-        /// For .NET Framework 4.7, the DLL must be placed in an architecture-specific subdirectory:
-        /// - x86/Yubico.NativeShims.dll for 32-bit processes
-        /// - x64/Yubico.NativeShims.dll for 64-bit processes
-        /// The correct version is loaded at runtime based on the process architecture.
-        /// </remarks>
-        internal const string NativeShims = "Yubico.NativeShims.dll";
-
-        /// <summary>
-        /// Ensures the native library is properly loaded for .NET Framework 4.7.
-        /// </summary>
-        /// <exception cref="DllNotFoundException">
-        /// Thrown when the native library cannot be loaded. This could be due to:
-        /// - Missing DLL file in the architecture-specific directory (x86/x64)
-        /// - Incorrect architecture (x86/x64 mismatch)
-        /// - Missing dependencies
-        /// - Insufficient permissions
-        /// </exception>
-        /// <remarks>
-        /// This method must be called before any P/Invoke calls are made.
-        /// The implementation details are handled in Libraries.Net47.cs.
-        /// </remarks>
-        public static void EnsureInitialized() => Net47Implementation.Initialize();
-#else
         /// <summary>
         /// The filename of the native shims library for modern .NET versions.
         /// </summary>
@@ -81,6 +55,6 @@ namespace Yubico.PlatformInterop
         /// This method exists only for API compatibility with .NET Framework 4.7 code.
         /// </remarks>
         public static void EnsureInitialized() { }
-#endif
     }
 }
+#endif
--- a/Yubico.Core/src/Yubico/PlatformInterop/Libraries.Net47.cs
+++ b/Yubico.Core/src/Yubico/PlatformInterop/Libraries.Net47.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #if NET47
+
 using System;
 using System.IO;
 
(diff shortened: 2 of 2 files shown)


Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
```diff
@@ -12,12 +12,48 @@
...
+using System.Runtime.InteropServices;
+#endif
+
 namespace Yubico.PlatformInterop
 {
     internal static partial class Libraries
     {
 #if NET47
-        internal const string NativeShims = "Yubico.NativeShims.dll";
+        internal const string NativeShims = "Yubico.NativeS hims.dll";
```
GregDomzalski, line 26: Extraneous space in `Shims`

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
```diff
@@ -12,12 +12,48 @@
...
+
+            _isNativeShimsIsLoaded = true;
+        }
+        
+        private static string NativeShimsPath => Path.Combine(
+            AppDomain.CurrentDomain.BaseDirectory,
+            Environment.Is64BitProcess ? "x64" : "x86",
+            NativeShims);
+
+        [DllImport("kernel32.dll", CharSet = CharSet.Auto, SetLastError = true)]
```
GregDomzalski, line 54: Look to see what we do with invoking other Win32 APIs with strings. There's the whole A/W variants. I think we probably want to make sure we're calling LoadLibraryW with CharSet = CharSet.Unicode...

DennisDyallo: Thanks for the input. I have revised the solution to make use of the [UnmanagedDynamicLibrary](https://github.com/Yubico/Yubico.NET.SDK/blob/develop/Yubico.Core/src/Yubico/PlatformInterop/UnmanagedDynamicLibrary.cs)

DennisDyallo (commented):
lgtm

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
```diff
@@ -12,14 +12,75 @@
...
+        /// - Incorrect architecture (x86/x64 mismatch)
+        /// - Missing dependencies
+        /// - Insufficient permissions
+        /// </exception>
+        /// <remarks>
+        /// This method must be called before any P/Invoke calls are made.
+        /// The implementation details are handled in Libraries.Net47.cs.
+        /// </remarks>
+        public static void EnsureInitialized() => Net47Implementation.Initialize();
 #else
```
GregDomzalski, line 65: Since this is the same partial class, why not just keep all NET47 stuff in the NET47 class? You could rename this file to Libraries.NetStandard.cs or something to make the separation more obvious.

Yubico.Core/src/Yubico/PlatformInterop/NativeMethods.cs
```diff
@@ -24,6 +24,11 @@ internal static partial class NativeMethods
         private const string MacDlLib = "libdl.dylib";
         private const string LinuxDlLib = "libdl.so";
 
+        static NativeMethods()
+        {
+            Libraries.EnsureInitialized();
+        }
+
```
GregDomzalski, line 31: Why not embed this in the `Libraries.Net47.cs` file? Do we need the level of indirection here? Couldn't we just make loading NativeShims part of the static constructor of the Libraries class?

```csharp
internal static Libraries()
{
   // Wrap in the try/catch like you already do. Omitted for brevity.
   if (_nativeShims != null)
   {
       return;
   }

   _nativeShims = UnmanagedDynamicLibrary.Open(NativeShimsPath);
}

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
GregDomzalski: Apparently I can't resolve my own comment threads? 😂

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
GregDomzalski: Good call! Like it.

GregDomzalski:
Just two final thoughts - otherwise looks good to me.

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
DennisDyallo: I gotcha

Yubico.Core/src/Yubico/PlatformInterop/Libraries.cs
DennisDyallo: Good catch

Yubico.Core/src/Yubico/PlatformInterop/NativeMethods.cs
DennisDyallo: I tried that first, but it wouldn't work. I did some research and it could be that the DllImport is executed at an earlier stage that static constructors. Then you might ask, well then, why does it work when you put the EnsureInitialized in the NativeMethods constructor? 
Answer is, I don't know. Could it be because of higher importance since the DllImport and the constructors are in the same class? I even tried throwing an exception in the Libraries constructor just to see if that would get thrown before the DllImport fails, which it didnt. So.. yeah. Weird one.

GregDomzalski: 😁 Yeah, I figured it may have been something like that. A mystery for another time perhaps. OK to resolve thread.

GregDomzalski:
OK, ok - two last foods for thought. Take it if you like the idea.

Instead of `#if !NET47` around the entire `Libraries.cs` file, let's scope it down to just the `internal const string` definition.

We can get rid of the `EnsureInitialized` method in the non NET47 file altogether, _if_ we put an `#if NET47` around the `static NativeMethods()` method decl in `NativeMethods.cs`.

I dunno. You're call. I think it's also fine as is. 😄

Consider this my sign-off.

PR #169, opened 2024-12-17
ci: added github attestation to build assets

DennisDyallo:
added github attestation to build assets, 

users will be able to prove that the assets were downloaded from github using 

gh attestation verify PATH/TO/YOUR/BUILD/ARTIFACT-BINARY -R Yubico/Yubico.NET.SDK


DennisDyallo:
fyi @dainnilsson

PR #171, opened 2024-12-17
ci: run full build periodically

DennisDyallo:
This will run builds periodically to ensure that we are always in a ready to release-state


DennisDyallo:
fyi @dainnilsson

PR #174, opened 2024-12-17
ci: fixed bug in sign.ps1

DennisDyallo:
fixed bug where the script attempted to veryfy the zips and not the nupkgs inside.


DennisDyallo:
@dainnilsson fyi

PR #204, opened 2025-03-05
docs: Cleared up confusion about challenge-response algorithms and features

equijano21:
# Description
Cleared up confusion about challenge-response algorithms and features in the documentation as discussed internally.

### DIFF
--- a/Yubico.YubiKey/docs/users-manual/application-otp/challenge-response.md
+++ b/Yubico.YubiKey/docs/users-manual/application-otp/challenge-response.md
@@ -20,9 +20,9 @@ limitations under the License. -->
 
 The other OTP application configurations ([Yubico OTP](xref:OtpYubicoOtp), [OATH HOTP](xref:OtpHotp),
 and [static password](xref:OtpStaticPassword)) require the user to activate the configured [slot](xref:OtpSlots) (by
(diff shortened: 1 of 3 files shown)


DennisDyallo (approved):
LGTM. Ill leave it to @joostd to follow up and do the final approve.

equijano21:
Thanks, Joost! It seems like this is really an issue of terminology. With HMAC-SHA1 challenge-response, you can use one of two challenge types: a value specified by the host application or the current Unix time. Regardless of the challenge type, you can receive the result as a byte array or numeric code. I think it makes sense to refer to the numeric codes as HOTPs and TOTPs, and we could call the byte array result from an application-specified challenge an "HMAC-SHA1 hash value". But what do we call the byte array result from a time-based challenge?

joostd:
> ... we could call the byte array result from an application-specified challenge an "HMAC-SHA1 hash value". But what do we call the byte array result from a time-based challenge?

The byte array response for application-specified challenges (whether HOTP or TOTP) are in fact HMAC-sHA1 values - they are just not truncated into numeric codes yet. So I think your suggestion for "HMAC-SHA1 hash value" is fine in both cases.

equijano21:
Closing this PR and opening a new one (https://github.com/Yubico/Yubico.NET.SDK/pull/221) due to major changes in doc file structure. All commits here included in new PR.

PR #208, opened 2025-03-19
fix: Stop send short APDUs when exceeded max APDU size

DennisDyallo:
# Description
This PR stops sending data to the YubiKey after it has exceeded the max allowed APDU size. 
Previously, the SDK would hammer the YubiKey until it had emptied its buffer.

It also adds the MaxApduSize constants.

The valid sizes (in bytes) for the smartcard connection on the YubiKey are as follows
NEO: 1390 
YK4: 2038 
YK4_3 : 3062

Fixes: YESDK-1329

## Type of change

- [x] Refactor (non-breaking change which improves code quality or performance)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

## How has this been tested?
Added unit tests in SmartCardTransformTests.cs

**Test configuration**:
* OS version: Arch Linux
* Firmware version: 5.4.3
* Yubikey model[^1]: 5 USBA

## Checklist:

- [x] My code follows the [style guidelines](https://raw.githubusercontent.com/Yubico/Yubico.NET.SDK/CONTRIBUTING.md) of this project 
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/CommandChainingTransform.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/CommandChainingTransform.cs
@@ -13,7 +13,10 @@
 // limitations under the License.
 
 using System;
+using System.Globalization;
+using Microsoft.Extensions.Logging;
 using Yubico.Core.Iso7816;
+using Yubico.Core.Logging;
 
 namespace Yubico.YubiKey.Pipelines
 {
@@ -23,7 +26,9 @@ namespace Yubico.YubiKey.Pipelines
     /// </summary>
     internal class CommandChainingTransform : IApduTransform
     {
-        public int MaxSize { get; internal set; } = 255;
+        private readonly ILogger _log = Log.GetLogger<CommandChainingTransform>();
+
+        public int MaxChunkSize { get; internal set; } = 255;
 
         readonly IApduTransform _pipeline;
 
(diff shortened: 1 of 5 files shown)


DennisDyallo:
@equijano21 With this PR I am adding a understandable exception to the user when they try to send a payload that's too large. 

This PR could go hand in hand with the documentation updates on the different `maxApduSizes` for the YubiKey variants

Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/CommandChainingTransform.cs
```diff
@@ -46,53 +47,77 @@ public ResponseApdu Invoke(CommandApdu command, Type commandType, Type responseT
...
+            return responseApdu;
+        }
+
+        private ResponseApdu FailedResponse(ResponseApdu responseApdu)
+        {
+            var currentCulture = CultureInfo.CurrentCulture;
+            string errorMessage = responseApdu.SW switch
+            {
+                SWConstants.WrongLength => string.Format(
+                    currentCulture, "Sent data exceeds max allowed length by YubiKey. (SW: 0x{0})",
```
dainnilsson, line 112: Technically, this might be false. There are other reasons we may get a WRONG_LENGTH 0x6700 SW from the YubiKey, so while the error *may* indicate that the amount of data sent exceeds the capacity, it's not guaranteed to be so.

dainnilsson: In the context of *any* APDU, generally speaking. It may be that the YubiKey specifically will always buffer the full request prior to parsing any of it, but that is not really a good assumption to make. Sure, that would allow you to infer that the issue is "too big" when there are still additional chunks to send, but that doesn't really help you when the final chuck is sent and you get that response (maybe the last fragment went over the limit, maybe it's just the response code from processing the command normally).

dainnilsson (approved):

Issue #182, opened 2024-12-25
[BUG] DeleteSlot / DeleteSlotConfiguration throws fault when not failing


virot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When calling DeleteSlot(Slot slot), it removes the configuration, but also returns an exception.
`System.InvalidOperationException: YubiKey Operation Failed [Warning, state of non-volatile memory is unchanged.]
                           at Yubico.YubiKey.Otp.Operations.DeleteSlotConfiguration.ExecuteOperation()
                           at Yubico.YubiKey.Otp.Operations.OperationBase`1.Execute()
                           at Yubico.YubiKey.Otp.OtpSession.DeleteSlot(Slot slot)`

  This is a bit similar to #48, but no COM code involved. I have reduced the issue to a small amount of code..
  https://github.com/virot/YubiKeyTests/blob/main/OTP_Remove_throws/Program.cs

If I enable the SDK Debugging code I see the following:
`info: Yubico.YubiKey.Oath.OathSession[0]
      Connecting to Otp
info: Yubico.YubiKey.Oath.OathSession[0]
      Connected to Otp
info: Yubico.Core.Devices.Hid.IHidConnection[0]
      Reading the OTP status.
info: Yubico.Core.Devices.Hid.WindowsHidDevice[0]
      Updating last used for HID:1050:0407 Usage:Keyboard \\?\HID#VID_1050&PID_0407&MI_00#8&57ad29&0&0000#{4d1e55b2-f16f-11cf-88cb-001111000030}\KBD to 01:27:46.8005131
Configuration state of ShortPress: False
warn: Yubico.YubiKey.Oath.OathSession[0]
      Exception caught when disposing Session: System.NotSupportedException, A YubiOTP APDU was received with instruction number 164. This instruction is not supported.`

So it seems that it is doing what should be done, removing the configuration then it throws a fault.

### Expected Behavior

It should remove the configuration and dont throw a fault.

### Steps To Reproduce

- [Clone / download](https://github.com/virot/YubiKeyTests/blob/main/OTP_Remove_throws/Program.cs)
- Build
- Run with a YubiKey (I have seen this complete on a FW 5.4.3, but then start to fail time after time.)

### Version

1.12.1

### Version

5.7.1

### Anything else?

Works on 5.4.3
![image](https://github.com/user-attachments/assets/52a2f514-5ea1-4c2e-be51-5d2d20c278f2)

[...]

PR #276, opened 2025-07-07
fix(otp): Handle case when last slot is deleted

DennisDyallo:
This pull request modifies the error handling logic in the `OtpErrorTransform` pipeline to improve clarity and handle edge cases more effectively. The changes include introducing a helper method for failed APDU responses and refining sequence validation logic.

Fixes: #182 

Error handling improvements:

* [`Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/OtpErrorTransform.cs`](diffhunk://#diff-10e4a6d21a6db46c24f1c2bb3090cc9d6bf1bc40b13aa048f3e2a1307b7ce832L65-R78): Added a `FailedApdu` helper method to streamline the creation of failed APDU responses. This simplifies the code and avoids redundancy.
* [`Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/OtpErrorTransform.cs`](diffhunk://#diff-10e4a6d21a6db46c24f1c2bb3090cc9d6bf1bc40b13aa048f3e2a1307b7ce832L65-R78): Updated the sequence validation logic to account for cases where the programming sequence is set to `0` when deleting the last slot, ensuring proper handling of this edge case.

### DIFF
--- a/.github/workflows/build-pull-requests.yml
+++ b/.github/workflows/build-pull-requests.yml
@@ -52,7 +52,8 @@ jobs:
         with:
           global-json-file: global.json
 
-      - name: Add local NuGet repository
+      # Add local NuGet repository so that we can use internal pre-release versions
+      - name: Add local NuGet repository for prerelease versions
         run: dotnet nuget add source --username ${{ github.actor }} --password ${{ secrets.GITHUB_TOKEN }} --store-password-in-clear-text --name github "https://nuget.pkg.github.com/Yubico/index.json"
 
       - name: Build Yubico.NET.SDK.sln
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,21 +80,23 @@ jobs:
       - uses: actions/setup-dotnet@v4
         with:
           global-json-file: "./global.json"
-      - name: Add local NuGet repository
+
+      # Add local NuGet repository if version is not a plain release (i.e., has a suffix, e.g., -prerelease)
+      - name: Add local NuGet repository for non-release versions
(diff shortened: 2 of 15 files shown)


Yubico.YubiKey/src/Yubico/YubiKey/Pipelines/OtpErrorTransform.cs
```diff
@@ -57,25 +58,36 @@ public ResponseApdu Invoke(CommandApdu command, Type commandType, Type responseT
...
+                var readStatusResponse = new ReadStatusResponse(responseApdu);
+                var otpStatus = readStatusResponse.GetData();
+                byte nextSequence = otpStatus.SequenceNumber;
 
                 // If we see the sequence number change, we can assume that the configuration was applied successfully. Otherwise
                 // we just invent an error in the response.
-                return afterSequence != expectedSequence
-                    ? new ResponseApdu(responseApdu.Data.ToArray(), SWConstants.WarningNvmUnchanged)
-                    : responseApdu;
+                return IsValidSequenceProgression(beforeSequence, nextSequence, otpStatus)
```
dainnilsson: Since you're passing `otpStatus` you don't really need to extract `nextSequence` above, you can just use `afterStatus.SequenceNumber` in the helper function. You could also use `afterStatus.ShortPressConfigured` and `afterStatus.LongPressConfigured` instead of the touch level and masking out the bits. The extra 3 bits are related to the specific configuration of the slots, but if both slot 1 and 2 are unconfigured, they *have* to be 0.

PR #278, opened 2025-07-10
docs: updated fido2 reset information

equijano21:
# Description

Updated the FIDO2 reset documentation:

- Fixed an error in the instructions (no need to send the reset command prior to rebooting the key)
- Clarified timeout durations (for initiating the reset after key reboot and touching the key to complete the reset)
- General improvements

Fixes: RFE-3296

## How has this been tested?

Local docs build.


DennisDyallo:
LGMT as well. Thanks @equijano21

Issue #191, opened 2025-01-26
[BUG] EnumerateRpsBeginResponse.GetData fails on some firmware versions when no RelyingParties exists


virot:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

On: YubiKey 5 NFC / 5.4.3
Security Key C by Yubico / 5.4.3

Causes:
Exception             : Yubico.YubiKey.Fido2.Ctap2DataException: The FIDO2 info returned is invalid.
                           at Yubico.YubiKey.Fido2.Commands.EnumerateRpsBeginResponse.GetData()
                           at Yubico.YubiKey.Fido2.Fido2Session.EnumerateRelyingParties()

When adding a single RP it starts working as intended

### Expected Behavior

Return zero same as 5.7.1 not a exception

### Steps To Reproduce

Sample software:
https://github.com/virot/YubiKeyTests/blob/main/fido2_empty_fido/Program.cs

This does not use a KeyCollector, my project powershellYK does, and both are affected the same way and I am to lazy for my test code

### Version

1.12.1

### Version

5.4.3

PR #316, opened 2025-09-25
fix: Fido2Session would incorrectly throw exceptions on older keys

DennisDyallo:
This pull request makes targeted improvements to the permission validation. The main focus is on clarifying when certain checks and actions should occur based on the requested permissions and authenticator capabilities.

Fixes: #191 

**Validation logic enhancements:**

* Refined the `ValidateParameters` method to only check for the presence of a relying party ID and the support for `pinUvAuthToken` when permissions are actually being requested, preventing incorrect exceptions and making the logic clearer.

**Authentication token handling:**

* Updated the `GetAuthToken` method to clarify that `AddPermissions` sets the auth token when `pinUvAuthToken` is supported, and switched to using `TryVerifyPin` to set the auth token when it is not supported, improving code readability and intent.

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/Fido2/Fido2Session.Pin.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/Fido2/Fido2Session.Pin.cs
@@ -375,11 +375,11 @@ namespace Yubico.YubiKey.Fido2
 
             if (AuthenticatorInfo.GetOptionValue(AuthenticatorOptions.pinUvAuthToken) == OptionValue.True)
             {
-                AddPermissions(permissions, relyingPartyId);
+                AddPermissions(permissions, relyingPartyId); // This will set the auth token
             }
             else
             {
-                AddPermissions(PinUvAuthTokenPermissions.None, null); // TODO use VerifyPin here instead directly!!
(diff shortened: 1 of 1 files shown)


Yubico.YubiKey/src/Yubico/YubiKey/Fido2/Fido2Session.Pin.cs
```diff
@@ -299,17 +299,18 @@ public void AddPermissions(PinUvAuthTokenPermissions permissions, string? relyin
...
             {
-                if (p == PinUvAuthTokenPermissions.None && rid is not null)
+                bool isRequestingPermissions = p != PinUvAuthTokenPermissions.None;
+                if (!isRequestingPermissions && rid is not null)
                 {
                     throw new ArgumentException(ExceptionMessages.Fido2PermsMissing);
                 }
 
-                if (AuthenticatorInfo.GetOptionValue(AuthenticatorOptions.pinUvAuthToken) != OptionValue.True)
+                if (isRequestingPermissions && AuthenticatorInfo.GetOptionValue(AuthenticatorOptions.pinUvAuthToken) != OptionValue.True)
```
DennisDyallo, line 308: Here is the fix

dainnilsson (approved):

PR #322, opened 2025-10-07
docs: minor edits to U2F and FIDO2 pages

equijano21:
# Description

Edited a product name ("YubiKey Security Key Series" -> "Security Key Series") and removed a statement that incorrectly described FIDO U2F behavior.

## How has this been tested?

Local docs build

### DIFF
--- a/docs/users-manual/application-fido2/how-fido2-works.md
+++ b/docs/users-manual/application-fido2/how-fido2-works.md
@@ -21,7 +21,6 @@ limitations under the License. -->
 At its foundation, FIDO2 is very similar to U2F. However, there are differences. For
 example:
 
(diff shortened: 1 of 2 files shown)


docs/users-manual/application-fido2/how-fido2-works.md
```diff
@@ -21,7 +21,6 @@ limitations under the License. -->
 At its foundation, FIDO2 is very similar to U2F. However, there are differences. For
 example:
 
-- In FIDO2 the private key is stored on the authenticator
```
equijano21, line 24: This statement makes it sounds like the private key is NOT stored on the authenticator with U2F, which is incorrect.

DennisDyallo (approved):

PR #340, opened 2025-11-07
docs: improved 1.14.1 release notes

equijano21:
# Description

Improved and significantly shortened the release notes for 1.14.1. The AI-generated version had way too much detail, including items that really shouldn't be release-noted and stuff that was actually wrong, and it made it difficult to understand what the big problems were that were resolved by the patch.

## How has this been tested?

Local docs build.


DennisDyallo (approved):
Agree. This is cutting to the chase. Thanks Elena!

PR #349, opened 2025-11-21
docs: third-party payment extension

equijano21:
# Description

User's manual documentation covering support for the third-party payment extension.

Also updated release notes and added a new TOC section for FIDO2 extensions.

## How has this been tested?

Local docs build. Tested sample code in Visual Studio with sample credential/assertion parameter data.

**Test configuration**:
* OS version: macOS 15.6.1 (Sequoia)
* Firmware version: 5.7.1
* YubiKey: 5C NFC

### DIFF
--- a/docs/users-manual/toc.yml
+++ b/docs/users-manual/toc.yml
@@ -339,7 +339,7 @@
               href: application-fido2/large-blobs.md
         - name: Additional extensions
           items:
-            - name: HMAC secret extension
+            - name: HMAC secret extensions
               href: application-fido2/hmac-secret.md
             - name: Third-party payment extension
               href: application-fido2/thirdpartypayment.md


docs/users-manual/toc.yml
```diff
@@ -337,8 +337,12 @@
               href: application-fido2/cred-blobs.md
             - name: Large blobs
               href: application-fido2/large-blobs.md
-        - name: HMAC secret extension
-          href: application-fido2/hmac-secret.md
+        - name: Extensions
```
DennisDyallo: hmm, adding a new Extensions node here. This kind of makes it look like these are the only two extensions. I believe the other extensions are woven into the above .md files (blobs, minpinlength, etc). Should we then write docs for each extension?

DennisDyallo: Would an interim solution work? Calling it "Additional extensions"? While we flesh it out?

equijano21: Good catch. I think "Additional extensions" makes perfect sense—let's go with that. :)

PR #356, opened 2025-12-01
docs: edits to page covering max PIV cert sizes 

equijano21:
# Description
Removed inaccurate statements about typical PIV cert sizes, reworked/streamlined chapter.

Addresses RFE-3560.

## How has this been tested?

Local docs build.


docs/users-manual/application-piv/cert-size.md
```diff
@@ -40,33 +34,19 @@ YubiKey.
...
-|    YubiKey Version    | Maximum Total Cert<br/>Space Available | Number of Certs<br/>at Size | Number of Certs<br/>at Maximum Size |
-|:---------------------:|:--------------------------------------:|:---------------------------:|:-----------------------------------:|
-| before 4.0 (e.g. NEO) |                  8100                  |    4 certs at 2025 bytes    |        4 certs at 2025 bytes        |
-|          4.x          |              about 49,800              |   24 certs at 2075 bytes    |       16 certs at 3052 bytes        |
-|       4.x FIPS        |              about 49,800              |   24 certs at 2075 bytes    |       16 certs at 3052 bytes        |
-|          5.x          |              about 50,000              |   24 certs at 2084 bytes    |       16 certs at 3052 bytes        |
-|       5.x FIPS        |              about 49,890              |   24 certs at 2079 bytes    |       16 certs at 3052 bytes        |
-
-Note that that total amount of storage on a YubiKey (for certs, for PUT DATA objects,
+Note that that total amount of storage on a YubiKey (for certs, PUT DATA objects,
```
JMarkstrom: The (original) text is confusing. The PIV applet storage is 50K (AFAIK), with about 1K(?) occupied by from the factory by the PIV attestation objects. This means available theoretical space is 49K unless the user deletes the attestation cert. Further, the statement: " there will be very little space left for anything else." suggests that there's something else the audience should put on the applet besides certificates. Very strange...

I would move this (shortened) to the opening paragraph. Actual storage should be verified by fw team.

The Yubikey PIV applet can store up to 24 certificates with corresponding private keys. In practice the number of certificates and keys that can be stored is determined by choice of algorithm and key length (e.g. RSA `1024` vs RSA `4096`), certificate complexity (e.g. use of OIDs, size attributes), the presence of PIV attestation objects etcetera.

**Note**: the total storage capacity of the YubiKey PIV applet is X with Y being occupied by the PIV attestation objects.

docs/users-manual/application-piv/cert-size.md
```diff
@@ -40,33 +34,19 @@ YubiKey.
...
-total size limit is reached, the YubiKey won't be able to store any more, even if some of
-the slots are empty.
-
-However, because a real world application will probably not use certs bigger than 2,000
-bytes, it is not likely it will ever run into a total space limitation and will be able
-to store up to 24 certs.
\ No newline at end of file
+there will be very little space left for anything else.
+
+|    YubiKey Version    | Maximum Total Cert<br/>Space Available | Maximum Average<br/>Cert Size | Number of Certs<br/>at Maximum Size |
```
JMarkstrom: Instead of "YubiKey Version", write: "Firmware"
Instead of "Number of Certs", write: "Number of certificates*"

Add "*" at table footer with text:

"*The approximate number of certificates with corresponding private keys"

docs/users-manual/application-piv/cert-size.md
```diff
@@ -40,33 +34,19 @@ YubiKey.
 
 ## Total space available for certificates
 
-Although a YubiKey 5.x will allow a 3052-byte cert in one of the slots, it will not be
-able to store 24 certs that big.
+Although YubiKeys with firmware version 4.x and higher will allow 3052-byte certificates, they will not be able to store 24 certificates of that size due to the key's total certificate space limit. Even if a YubiKey has empty certificate slots available, you cannot fill them once the maximum certificate space has been reached.  
 
-A NEO (pre-4.0), only has four slots, and will be able to hold four certs of the maximum
+However, a YubiKey NEO, which only has four slots, will be able to hold four certs of the maximum
```
JMarkstrom: NEO is long dead. No need to confuse the audience with it (IMHO).

docs/users-manual/application-piv/cert-size.md
```diff
@@ -40,33 +34,19 @@ YubiKey.
 
 ## Total space available for certificates
 
-Although a YubiKey 5.x will allow a 3052-byte cert in one of the slots, it will not be
-able to store 24 certs that big.
+Although YubiKeys with firmware version 4.x and higher will allow 3052-byte certificates, they will not be able to store 24 certificates of that size due to the key's total certificate space limit. Even if a YubiKey has empty certificate slots available, you cannot fill them once the maximum certificate space has been reached.
```
JMarkstrom: Skip 4.x mention. I don't know if we should fixate on "3052" like this. The paragraph also seems redundant considering the previous statements.

docs/users-manual/application-piv/cert-size.md
```diff
@@ -18,18 +18,12 @@ limitations under the License. -->
...
-over 1,000 bytes, but rarely over 2,000. It is unlikely that you will run into limitations
-on the YubiKey.
-
-Nonetheless, these are the space limitations for certs in the PIV application on the
-YubiKey.
+It is possible to store up to 24 private key/certificate pairs in the PIV slots for YubiKeys with firmware version 4.x and higher. However, there are limits to the size of each certificate and the total space available for all certificates.
 
 ## Maximum size for a single certificate
 
+If you attempt to load a certificate that is larger than the key's maximum allowable certificate size (as indicated in the table below), the YubiKey will reject it, and the SDK will throw an exception.
```
JMarkstrom: I don't think this is a correct statement. If you want to keep it, I would use "YubiKey's" instead of "key's" (otherwise its confusing since we are talking about certificates and private, public keys). Does this table indicate certificate maximum size or is it "private key + certificate" combined maximum size?

docs/users-manual/application-piv/cert-size.md
```diff
@@ -18,18 +18,12 @@ limitations under the License. -->
...
-It is possible to store up to 24 private key/certificate pairs in the PIV slots. However,
-there are space limitations.
-
-In the real world, certificates are generally less than 1,000 bytes. Some large certs are
-over 1,000 bytes, but rarely over 2,000. It is unlikely that you will run into limitations
-on the YubiKey.
-
-Nonetheless, these are the space limitations for certs in the PIV application on the
-YubiKey.
+It is possible to store up to 24 private key/certificate pairs in the PIV slots for YubiKeys with firmware version 4.x and higher. However, there are limits to the size of each certificate and the total space available for all certificates.
```
JMarkstrom: I would skip any mention of YK fw. 4 (it was released in 2015 and we cannot reasonably expect customers to enroll certificates on YK 4 ten years later, right?). Maybe something along the lines of: 

> The Yubikey PIV applet can store up to 24 certificates with corresponding private keys. In practice the number of certificates and keys that can be stored is determined by choice of algorithm and key length (e.g. RSA `1024` vs RSA `4096`), certificate complexity (e.g. use of OIDs, size attributes), the presence of PIV attestation objects etcetera.

docs/users-manual/application-piv/cert-size.md
equijano21: I don't know about the table data—I wasn't the original writer of this page. I can verify the information with the firmware team if you're not sure it's correct.

Also, this statement was already in the docs, I just relocated it and reworked the sentence a bit.

docs/users-manual/application-piv/cert-size.md
equijano21: I don't see a reason to remove mention of the firmware version—all we're saying here is that X is true for YubiKeys with firmware Y and above. That doesn't read to me as something that would confuse a developer with 5.x keys.

docs/users-manual/application-piv/cert-size.md
equijano21: I think the diff might be confusing—here's what the HTML output looks like:

Screenshot 2025-12-02 at 10 00 55 AM

equijano21: What do you mean by "fixate on 3052"? This page just covers max cert size and total cert space available, so the 3052 max byte size is kind of a critical piece of info, especially given the contents of the second table.

docs/users-manual/application-piv/cert-size.md
equijano21: I'm really hesitant to remove product information like this. If we want to start scrubbing any mention of NEO keys from the docs, I'd like to get eng and product consensus on it first.

docs/users-manual/application-piv/cert-size.md
equijano21: Added note about cert sizes in practice: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/d3bd2aed96859a2ea224ebfd25ba28e2f2212d30

docs/users-manual/application-piv/cert-size.md
equijano21: Spelled out "certificates" across tables and other page text: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/56e0ec297bd360ed8f8088a040c07b5135f5362f

equijano21: Wrote out model and firmware in tables, rearranged as necessary: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/485ca637b6c9702a7d79bc4801a182ef87b4f857

(model is important because there are differences in total size available between fips and standard keys)

docs/users-manual/application-piv/cert-size.md
equijano21: Wrote out "YubiKey" where necessary: https://github.com/Yubico/Yubico.NET.SDK/commit/fefe4c07c462d0cf4598aea3bde350528539a580

equijano21:
@JMarkstrom I made a couple other changes after discussing these docs with the firmware team: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/38c5d2b9a4f985b861232c9a2f7c60a9e7e1f1f4

The PR is now ready for another review.

docs/users-manual/application-piv/cert-size.md
equijano21: Confirmed with the firmware team that the table data is correct.

docs/users-manual/application-piv/cert-size.md
equijano21: Additional updates regarding total storage space: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/38c5d2b9a4f985b861232c9a2f7c60a9e7e1f1f4

docs/users-manual/application-piv/cert-size.md
JMarkstrom: OK, up to you. I would at least spell out "YubiKey NEO" then, and "certificates" instead of "certs" (feels too colloquial to me).

docs/users-manual/application-piv/cert-size.md
JMarkstrom: Ok, I realize this is scope creep, but for starters the summary is _longer_ than the article. That does not make any sense. In fact the article is Soo short that it does not need any summary (I would remove it). Now, if you read the summary only then '3052' is a totally out of the blue number. You have to carefully read the table to make sense of the summary, and that is NOT a good approach to a summary. If you really want to keep the summary then I would suggest you put '3052' in some kind of context, e.g. referring to it there as the maximum size of an importable object.

docs/users-manual/application-piv/cert-size.md
equijano21: Updated product names in both tables: https://github.com/Yubico/Yubico.NET.SDK/pull/356/commits/de90d2d82a9aa46fca6dfd2f0083841714e763b5

PR #418, opened 2026-02-24
fix: don't retain ghost YubiKey devices after e.g. FIDO reset

DennisDyallo:
This pull request addresses an issue where "ghost" YubiKey device entries with incomplete information (specifically, `SerialNumber=null` and `FirmwareVersion=0.0.0`) were being retained in the device cache after a FIDO reset. These ghost entries could cause downstream errors and confusion for consumers of the device API. The changes add logic to detect and discard such incomplete device entries, preventing them from being cached or surfaced to consumers.

Fixes: `yesdk-1556`

The most important changes are:

**Device Cache Robustness:**

* Added a static `Default` instance (`0.0.0`) to the `FirmwareVersion` class for easier comparison when detecting incomplete device info.
* Updated the `YubiKeyDeviceListener.Update()` method to detect and skip devices with both `SerialNumber=null` and `FirmwareVersion=FirmwareVersion.Default`, logging a warning and preventing these incomplete entries from being cached.

**Documentation and Analysis:**

* Added a detailed markdown document describing the ghost device issue, steps to reproduce, root cause analysis, and the rationale for the implemented fix.


DennisDyallo:
@claude Can you take a look at this? Code review, correctness, intentional result, actual result, reliability, etc.

PR #452, opened 2026-04-13
test: SCP03 command chaining regression test (YESDK-1260)

DennisDyallo:
## Summary

- Adds integration test `Scp03_Piv_RSA2048Sign_WithCommandChaining_Succeeds` covering the scenario reported in YESDK-1260: RSA 2048 signing over an SCP03 channel, which triggers transport-level command chaining
- Cleans up stale comments in `ScpConnection.CreateScpPipeline`
- **No functional code change** — investigation confirmed the original pipeline order (`SCP03 → CommandChaining(255)`) is already correct

## Background

YESDK-1260 reported that `SW 6982` was returned for RSA signing over SCP03 while ECC signing worked fine. ECC doesn't trigger command chaining (small payload); RSA 2048 does (256-byte PKCS#1 block).

The correct pipeline order is:
1. **SCP03 encrypts + MACs the full plaintext APDU** (output can exceed 255 bytes)
2. **CommandChaining(255) splits the encrypted blob** into ≤255-byte transport chunks with `CLA | 0x10`
3. The YubiKey reassembles transport chunks before SCP03 decryption

This matches yubikey-manager's `ScpProcessor` implementation exactly. An earlier attempted fix that inverted this order (chunk first, then encrypt each chunk independently) was incorrect and has been reverted.

## Test plan

- [ ] Run `dotnet test Yubico.YubiKey/tests/integration/Yubico.YubiKey.IntegrationTests.csproj --filter "Scp03_Piv_RSA2048Sign_WithCommandChaining_Succeeds"` on a YubiKey 5 USB (non-FIPS)
- [ ] Confirm all 17 SCP03 integration tests still pass
- [ ] Confirm all unit tests pass (`dotnet test Yubico.YubiKey/tests/unit/Yubico.YubiKey.UnitTests.csproj`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Yubico.YubiKey/src/Yubico/YubiKey/Scp/ScpConnection.cs
+++ b/Yubico.YubiKey/src/Yubico/YubiKey/Scp/ScpConnection.cs
@@ -33,6 +33,7 @@ namespace Yubico.YubiKey.Scp
             : base(smartCardDevice, application, null)
         {
             var scpPipeline = CreateScpPipeline(keyParameters);
+
             var withErrorHandling = CreateParentPipeline(scpPipeline, application);
 
             // Have the base class use the new error augmented pipeline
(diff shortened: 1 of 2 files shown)


oskardanielsson-yubico (approved):
looks good

DennisDyallo:
thanks @oskardanielsson-yubico

PR #445, opened 2026-04-01
fix: High idle CPU in RDS/Terminal Server environments (SCARD_E_INVALID_HANDLE)

DennisDyallo:
## Summary

Closes #434. Also resolves #37 (same root cause, reported 2023-02-13 — the `SCARD_E_INVALID_HANDLE` gap in `UpdateContextIfNonCritical` survived every listener refactor since then).

When an RDS/Windows 365/Terminal Server session is disconnected and reconnected, the Windows Smart Card Service invalidates existing `SCARDCONTEXT` handles. `DesktopSmartCardDeviceListener` was calling `SCardGetStatusChange` in a tight loop with the stale handle. `WinSCard.dll` internally raises and unwinds a C++ exception (`CxxThrowException`) for every such call — confirmed in 6 of 10 minidumps. This ran thousands of times per second, pegging one CPU core.

The error code returned to managed code (`SCARD_E_INVALID_HANDLE = 0x80100003`) was not handled by `UpdateContextIfNonCritical`. The code logged the error and immediately retried with no backoff and no context re-establishment.

## Changes

- **`ISCardInterop` interface + `SCardInterop` concrete class** — extracts the four SCard P/Invoke calls behind an injectable interface, enabling cross-platform unit tests without hardware or Windows
- **`UpdateContextIfNonCritical`** — adds `SCARD_E_INVALID_HANDLE`, `SCARD_E_SYSTEM_CANCELLED`, `ERROR_BROKEN_PIPE` to the recovery switch (all three surface in RDS session transitions)
- **`Thread.Sleep(1000)` backoff** — applied after every recovery attempt to prevent a secondary tight loop if `SCardEstablishContext` also fails while the service is transitioning
- **`UpdateCurrentContext` fix** — reordered to dispose the old context BEFORE calling `EstablishContext`, preventing a handle-value recycling bug where Windows could reuse the freed IntPtr value for the new context, and the subsequent old `SafeHandle.Dispose()` would release the new context via `SCardReleaseContext`
- **Default path backoff** — unrecognised error codes also sleep 1000 ms (catch-all for unknown persistent errors)
- **CI-safe CPU test threshold** — raised from 500ms to 1500ms to tolerate concurrent test activity while still clearly catching the unfixed behavior (≥2500ms)
- **Robust recovery test assertion** — replaced fragile IntPtr comparison with `IsClosed` check, since handle value recycling is legitimate OS behavior

### Cross-platform mock tests (run on any CI platform, no hardware needed)
```powershell
[...]

PR #459, opened 2026-04-21
Fix smart card listener timeout handling and add backoff for immediate failures

TimLico:
# Description
This change fixes two smart-card listener issues in the .NET SDK. 
First, it treats SCARD_E_TIMEOUT as a normal polling outcome and exits the current update cycle without continuing into change detection or firing events. 
Second, it adds a backoff path for immediate non-timeout failures, so the listener does not hot-loop and flood logs when SCardGetStatusChange returns immediately with errors such as Result = 6.

The idea is to only slows the bad failure path
it preserves fast success handling and it preserves normal timeout behavior

timeout → Continue
cancelled → Stop
immediate failure → Backoff
success → keep processing normally.

**Issue fixed & Motivation and context:**
Fixes the smart-card listener log flooding issue where SCardGetStatusChange failures could be logged at sub-millisecond intervals instead of following the intended 100 ms polling cadence. (a simpler fix of just adding Thread.Sleep(CheckForChangesWaitTime) after the immediate failure path. That would likely reduce the log flooding with a small code change, but it felt more like a mitigation than a clean fix. The richer return approach seemed better because it makes the control flow explicit: CheckForUpdates() can tell the caller whether to continue, stop, or back off, and ListenForReaderChanges() can handle each case appropriately.)

**Dependencies:**
No external dependencies. PCS will need to update to the fixed .NET SDK package once the new NuGet release is published.

Fixes: # <[PROD-5297](https://yubico.atlassian.net/browse/PROD-5297)>

## Type of change

- [ ] Refactor (non-breaking change which improves code quality or performance)

## How has this been tested?

Tested by running PCS and comparing the logs.

## Checklist:

- [x] My code follows the [style guidelines](https://raw.githubusercontent.com/Yubico/Yubico.NET.SDK/CONTRIBUTING.md) of this project 
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/Yubico.Core/src/Yubico/Core/Devices/SmartCard/DesktopSmartCardDeviceListener.cs
+++ b/Yubico.Core/src/Yubico/Core/Devices/SmartCard/DesktopSmartCardDeviceListener.cs
@@ -29,6 +29,13 @@ namespace Yubico.Core.Devices.SmartCard
     /// </summary>
     internal class DesktopSmartCardDeviceListener : SmartCardDeviceListener
     {
+        private enum CheckForUpdatesResult
+        {
+            Continue,
+            Stop,
+            Backoff
+        }
+
         private static readonly string[] readerNames = new[] { "\\\\?\\Pnp\\Notifications" };
         private readonly ILogger _log = Logging.Log.GetLogger<DesktopSmartCardDeviceListener>();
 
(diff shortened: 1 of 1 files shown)


DennisDyallo:
Relates to #445

PR #474, opened 2026-04-29
build(core): float NativeShims version with packages.lock.json

DennisDyallo:
## Summary
- Replace hard-pinned `Yubico.NativeShims` version in `Yubico.Core.csproj` with floating range `1.*-*`.
- Enable `RestorePackagesWithLockFile` so each restore captures the resolved version in `packages.lock.json`.
- Stops needing a csproj bump every NativeShims prerelease publish, while keeping builds reproducible (lockfile pins the concrete version until an explicit `dotnet restore --force-evaluate`).

## Why
Currently every NativeShims prerelease forces a hand-edit of `Yubico.Core.csproj`. The floating-version + lockfile pattern is the standard NuGet way to "always get the latest on intent, never on accident":

- **Floating version (`1.*-*`)** — restore picks the highest 1.x including prereleases.
- **`packages.lock.json`** — captures the exact version resolved, committed to git. Subsequent restores honor the lock; only `--force-evaluate` re-floats.
- Net effect: no more drive-by version bumps, but no surprise upgrades either.

## Local verification
- `dotnet restore Yubico.Core/src/Yubico.Core.csproj` → generates lockfile, resolves to `1.16.1-prerelease.20260428.1` (same as the previous hard pin).
- `dotnet build Yubico.Core/src/Yubico.Core.csproj` → 0 warnings, 0 errors across `net472`, `netstandard2.0`, `netstandard2.1`.
- `dotnet build Yubico.YubiKey/src/Yubico.YubiKey.csproj` (downstream consumer) → clean.
- `dotnet test Yubico.Core/tests/Yubico.Core.UnitTests.csproj` → **525 passed, 0 failed, 21 skipped** (skips are platform-gated, expected on macOS).

## Test plan
- [ ] CI build succeeds on all platforms.
- [ ] CI restore produces the same resolved NativeShims version as the lockfile.
- [ ] Confirm with maintainers whether to add `--locked-mode` to CI in a follow-up (would fail CI if the lockfile drifts from a fresh resolve).

## Follow-ups (not in this PR)
- Consider applying the same pattern to `Yubico.YubiKey.csproj` and test projects if they grow direct floating dependencies.
- Add `--locked-mode` (or `<RestoreLockedMode>true</RestoreLockedMode>` under `Condition="'\$(GITHUB_ACTIONS)' == 'true'"`) so CI can never silently float past the lockfile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,8 +52,41 @@ permissions:
   contents: read
 
 jobs:
+  enforce-branch-policy:
+    name: Enforce branch policy
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+        with:
+          persist-credentials: false
+
+      # main branch is release-only — no prerelease Yubico packages allowed in the lockfile.
+      # Re-pin via `dotnet restore Yubico.Core/src/Yubico.Core.csproj --force-evaluate` against
+      # an "nuget.org-only" environment before merging develop -> main.
+      - name: Disallow prerelease Yubico.NativeShims on main
+        if: github.ref == 'refs/heads/main'
+        shell: bash
+        run: |
+          LOCK=Yubico.Core/src/packages.lock.json
+          # The package name and its "resolved" field live on separate lines, so use -A3 to
+          # bring the resolved line into the match window before checking for -prerelease.
+          if grep -A3 '"Yubico\.NativeShims"' "$LOCK" | grep -E '"resolved":.*-prerelease' >/dev/null; then
+            echo "::error file=$LOCK::main builds disallow prerelease Yubico.NativeShims. Re-pin the lockfile to a stable version (dotnet restore --force-evaluate against nuget.org) before merging."
+            grep -A3 '"Yubico\.NativeShims"' "$LOCK" | head -20
+            exit 1
+          fi
+          echo "Lockfile policy OK: no prerelease Yubico.NativeShims pinned."
+
   run-tests:
     name: Run tests
+    needs: enforce-branch-policy
     # Requires write permissions to publish test results
     permissions:
       checks: write         # Required to create check runs for test results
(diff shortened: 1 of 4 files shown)


DennisDyallo:
Bundled the branch-conditional NuGet policy into this PR (per discussion):

**Workflow changes (`.github/workflows/build.yml`)**
- New `enforce-branch-policy` job runs first; on `refs/heads/main` it greps `Yubico.Core/src/packages.lock.json` and **hard-fails** if a prerelease `Yubico.NativeShims` is pinned. `run-tests` and `build-artifacts` `needs:` this job so the policy failure surfaces with a clear actionable error before any restore attempt.
- `build-artifacts` now has two mutually exclusive `setup-dotnet` steps: non-main builds get the internal Yubico GitHub Packages feed (prereleases resolvable); main builds get nuget.org only.

**csproj change (`Yubico.Core/src/Yubico.Core.csproj`)**
- Added `<RestoreLockedMode Condition="'$(GITHUB_ACTIONS)' == 'true'">true</RestoreLockedMode>` so CI fails restore on any lockfile drift instead of silently re-floating.

**Implied release ritual**: before merging develop → main, re-pin the lockfile to a stable version:
```bash
dotnet restore Yubico.Core/src/Yubico.Core.csproj --force-evaluate
git commit -am "chore(core): repin Yubico.NativeShims to stable for release"
```
The new `enforce-branch-policy` guard makes that requirement explicit (with a pointer to the fix) instead of letting it surface as a cryptic restore-failure deep in the matrix.

Verified locally: YAML parses, `GITHUB_ACTIONS=true dotnet restore` succeeds, guard rehearsed against both current (fires) and synthesized stable (passes) lockfiles.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

DennisDyallo:
**CI failure fix pushed (commit `1eb905a2`)**

The Windows test job failed with NU1004:
> The package references have changed for net472. Lock file's package references: ... Microsoft.Bcl.HashCode:[6.0.0, ), ... Microsoft.NETFramework.ReferenceAssemblies:[1.0.3, ), ... project's package references: ... (no Microsoft.NETFramework.ReferenceAssemblies, no Microsoft.Bcl.HashCode) ...
> The packages lock file is inconsistent with the project dependencies so restore can't be run in locked mode.

**Root cause:** NuGet lockfiles are not OS-portable when implicit packages differ. `Microsoft.NETFramework.ReferenceAssemblies` is implicitly added by the .NET SDK when targeting `net472` on **non-Windows** (Mac/Linux don't have the framework reference assemblies installed locally; Windows does). The lockfile I committed was generated on macOS, so it includes that implicit package; on Windows CI, the project graph doesn't include it, and `<RestoreLockedMode>` hard-fails.

**Fix:** Dropped `<RestoreLockedMode>`. Replaced with an explanatory comment in the csproj. The other three pieces of the PR are unaffected and still satisfy all the original goals:

| Requirement | Mechanism | Status |
|---|---|---|
| Stop hand-bumping NativeShims version | `Version="1.*-*"` | ✅ kept |
| Reproducible restores | `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` + committed `packages.lock.json` | ✅ kept |
| No prereleases on main | `enforce-branch-policy` job in `build.yml` (greps lockfile, hard-fails on main) | ✅ kept |
| Branch-conditional NuGet sources | Two `setup-dotnet` steps in `build-artifacts` | ✅ kept |
| Fail CI on lockfile drift | `<RestoreLockedMode>` | ❌ **removed — see follow-up** |

**What we lose:** if someone adds a new `<PackageReference>` and forgets to commit the updated `packages.lock.json`, CI won't hard-fail — it'll silently update the in-memory lock and proceed. Acceptable for now.

**Follow-up to safely re-introduce locked mode:**
1. Either generate per-OS lockfiles on every CI matrix runner and commit each, OR
2. Add explicit `<PackageReference>` entries for every implicit root (`Microsoft.NETFramework.ReferenceAssemblies` for net472, etc.) with `PrivateAssets="All"`, so all OSes resolve identically.

Both are non-trivial and deserve their own PR. Tracking note added to `Plans/` (locally) for follow-up.

Verified locally: `GITHUB_ACTIONS=true dotnet restore` now succeeds with no lockfile delta; `Yubico.Core` builds clean across net472, netstandard2.0, netstandard2.1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #531, opened 2026-07-15
fix: align Yubico.NativeShims to 16 KB pages for Android 15+

maxbrousseau:
## Summary

The Linux `x86_64` build of `libYubico.NativeShims.so` is linked with **4 KB (`2**12`) segment alignment**, which fails the [Android 16 KB page-size requirement](https://developer.android.com/guide/practices/page-sizes) for apps running on Android 15+ devices that use 16 KB memory pages (and is now required by Google Play). The `arm64` build is unaffected because the aarch64 linker already defaults to 64 KB alignment.

This surfaced downstream: a .NET Android app that transitively bundles this shim fails its `validate-16kb-compatibility` gate on `lib/x86_64/libYubico.NativeShims.so: UNALIGNED (2**12)`.

## Change

Add a single linker option in the Linux branch of `Yubico.NativeShims/CMakeLists.txt`:

```cmake
add_link_options("-Wl,-z,max-page-size=16384")
```

- Forces 16 KB-aligned `LOAD` segments on both Linux architectures.
- No-op for `arm64` (already ≥16 KB); fixes `x86_64`.
- `zig cc` (used by the build scripts) forwards `-Wl,…` to LLD, which honors `-z max-page-size=`.

## Verification

I have **not** built the artifacts locally (no Linux toolchain on hand), so the numbers below are the *expected* result to confirm against the CI build output rather than a measured run.

Check the `LOAD` segment alignment of the produced `.so`:

```
readelf -lW linux-x64/libYubico.NativeShims.so | awk '$1=="LOAD"{print $NF}'
```

Expected: every `LOAD` segment reports alignment `0x4000` (16384) or larger. Prior to this change the `x86_64` build reports `0x1000` (4 KB).

If useful, I'm happy to add an assertion to the `build-linux-*.sh` scripts (fail if any `LOAD` alignment `< 0x4000`), mirroring the existing glibc-version check, so this can't silently regress — let me know if you'd like it folded into this PR.

### DIFF
--- a/Yubico.NativeShims/CMakeLists.txt
+++ b/Yubico.NativeShims/CMakeLists.txt
@@ -35,6 +35,10 @@ if (APPLE OR UNIX)
         set(BACKEND "pcsc")
         # RELRO + NOW = full read-only relocations (exploit mitigation)
         add_link_options("-Wl,-z,relro,-z,now,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/exports.gnu")
+        # 16 KB max page size: required for Android 15+ devices with 16 KB memory pages
+        # (and by Google Play). x86_64 links at 4 KB by default; arm64 already defaults to 64 KB.
+        # https://developer.android.com/guide/practices/page-sizes
+        add_link_options("-Wl,-z,max-page-size=16384")
     endif()
 
     # --- macOS / Linux: security hardening (GCC/Clang) ---


DennisDyallo:
Hi @maxbrousseau and thanks for deciding to contribute! We'll look into this shortly :)

DennisDyallo:
Hi @maxbrousseau — thank you again for this contribution and for the really thorough write-up. Your investigation correctly pinpointed the root cause (the x86_64 shim linking at 4 KB and failing the Android 15+ 16 KB page-size requirement), and it put us on exactly the right track. 🙏

After digging in a bit further, we decided to go in a slightly different direction and land the fix as **64 KB** (`-Wl,-z,max-page-size=65536`) rather than 16 KB, opened as #536. The motivation is future-proofing without regressing anything:

- `aarch64` already defaults to **64 KB** in LLD, so `max-page-size=16384` would have *lowered* the arm64 shim from 64 KB to 16 KB — which can break loading on 64 KB-page aarch64 Linux distros (e.g. RHEL/Fedora), a platform we support.
- **64 KB is a superset alignment**: it loads correctly on 4 KB, 16 KB *and* 64 KB page systems, so it fixes Android x86_64 while keeping arm64 safe. Raising alignment is backward/forward-compatible for loading; only lowering it below a target's runtime page size is a hazard.

We also added a small CI regression guard (`tests/check_alignment.sh`, wired into the native-shims workflow for both Linux arches) so this can't silently regress — very much in the spirit of the assertion you offered to add.

Credit to you for the diagnosis and direction. **Would #536 work for your use case?** If your downstream `validate-16kb-compatibility` gate is happy with 64 KB-aligned segments (it should be — 64 KB ≥ 16 KB), we think this covers you and then some. And please do keep contributing — this was a great catch and exactly the kind of issue we're glad to have surfaced. Thanks again!

maxbrousseau:
Thanks @DennisDyallo  — really appreciate you taking the time to dig into this and land the better 64 KB fix. #536 works for us, so I'll close this in favor of it. Thanks again!

DennisDyallo:
@maxbrousseau Hey I'm curious what you guys are building anyway? Seeing as building Android apps using this SDK (v1 SDK) is not specifically supported. Care to share?

PR #571, opened 2026-08-19
test(core): align ownership assertions with grouped leases

DennisDyallo:
## Summary

- update grouped-lease contention assertions to accept any held member of a physical YubiKey instead of pinning the requested transport
- require the one-YubiKey/one-connection diagnostic contract and a concrete held member ID
- align Core, FIDO HID, PIV, OATH, and YubiOTP test descriptions with device-wide ownership terminology

## Root cause

Some tests and comments predated device-wide connection leases and assumed that contention must identify the interface requested by the second caller. Grouped connections now claim all known member interfaces in ordinal order, so acquisition can first encounter any held member. For example, a SmartCard request can report a held `hid:` member. The runtime refusal is correct; the transport-specific assertions and per-interface descriptions were stale.

The audit found no stale functional assertion in Fido2 or WebAuthn. Their availability-based transport selection and no-fallback-on-failure tests match the current contract.

## Verification

- `dotnet toolchain.cs -- test --project Core` (866 passed, 3 expected skips)
- `dotnet toolchain.cs -- test --integration --project Piv --filter "FullyQualifiedName~ConnectAsync_SecondSmartCardConnection_WhilePivSessionOpen_IsRefused"` (hardware pass on connected YubiKeys 103 and 125)
- `dotnet toolchain.cs -- test --integration --project Piv --smoke` (167 unit tests and 76 hardware integration tests passed before the broader ownership-comment sweep)

### DIFF
--- a/src/Core/tests/Yubico.YubiKit.Core.IntegrationTests/Devices/FidoHidOwnershipIntegrationTests.cs
+++ b/src/Core/tests/Yubico.YubiKit.Core.IntegrationTests/Devices/FidoHidOwnershipIntegrationTests.cs
@@ -22,7 +22,7 @@ using Yubico.YubiKit.Tests.Shared.Infrastructure;
 namespace Yubico.YubiKit.Core.IntegrationTests.Devices;
 
 /// <summary>
-///     Hardware-gated confirmation of the exclusive FIDO HID interface ownership contract.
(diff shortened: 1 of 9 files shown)


DennisDyallo:
ok

PR #578, opened 2026-08-21
feat(aot): Native AOT support for v2 SDK libraries

### LINKED ISSUE #60: Yubico.NativeShims doesn't build as expected with AOT
When building an executable the native AOT compiler, Yubico.NativeShims.dll is output next to the executable instead of being trimmed away. There are also no compiler warnings about dependencies having trimming issues, which is odd. But I suspect it has to do with native shims being pre-built/native or shared vs static, it's not IL. It probably needs to be IL or static, or maybe static with a p/invoke wrapper class, in order for trimming to work (not sure haven't done one of these myself before).

## Expected output
single executable with no libraries around it

## What's actually happening
Yubico.NativeShims.dll in the output dir.

## Steps to reproduce
- `dotnet new console`
- `dotnet add package Yubico.YubiKey`
- Add `<PublishAot>true</PublishAot>` to the project file
- `dotnet publish -o ./out`

## Sample with expected output
Instead of adding the YubiKey reference add one to `Newtonsoft.Json`
You should get a ton of warnings about trimming, but the output is as expected absent of any Newtonsoft libs.

## Why do I want this?
if you take a peek at my latest builds: https://github.com/SheepReaper/yk-csr-generator/releases/tag/v1.0.0 I'm building 80MB assemblies because of the runtime bits included. Granted, if I didn't do single file, it would be way smaller, but then the user needs to have a runtime installed. I just want to build a single executable with no external dependencies, so I'm fine with the bloat.
However, in testing AOT builds, I was able to achieve a 1.5KB binary plus the Shim lib next to it with almost 3MB. Which is super nice, but we're back at the original problem of external dependencies.

## Additional references
- native static library: https://github.com/dotnet/samples/blob/main/core/nativeaot/NativeLibrary/README.md
- aot compatible library limitations: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/aot/native-aot-tutorial?view=aspnetcore-8.0&tabs=netcore-cli#libraries-and-native-aot
- declaring your libs to be aot compatible enables the aot-compatibility analyzers (compiler warnings): https://learn.microsoft.com/en-us/dotnet/core/deploying/native-aot/?tabs=net8plus%2Cwindows#aot-compatibility-analyzers

I don't know enough about the Shims lib to determine what about it is preventing AOT from working nicely, but worst case it's hitting one of the many limitations of AOT compatibility, which might be too much work to overcome. In which case, I'll be sticking to my thicc binaries...

DennisDyallo:
## Summary

YESDK: [YESDK-1594](https://yubico.atlassian.net/browse/YESDK-1594)

- Enables Native AOT compatibility metadata and analyzers for the ten v2 SDK library packages.
- Adds an internal Native AOT verification host that links reachable entry types from every supported library and runs the ordinary Core discovery entry point.
- Adds a recurring macOS Apple Silicon publish-and-run workflow.
- Defines an evidence-based support contract that distinguishes analyzer, link, runtime, hardware, and recurring continuous-integration coverage.

## Evidence boundaries

- All ten libraries are analyzer-checked and link-verified.
- Core discovery has run against physical YubiKeys on macOS, Windows, and Linux.
- Management device-information reads, a PIV session, and device monitoring have run under Native AOT on macOS.
- The remaining applet modules have link evidence but no representative protocol runtime evidence under Native AOT.
- Recurring coverage is macOS Apple Silicon only and runs without hardware. Its empty discovery result proves execution of the Core entry point, not that every native transport loaded.

The normal build validates the exact ten supported project opt-ins, verification-host project references, and linked entry-type anchors.

## Native dependencies

The released `Yubico.NativeShims` package uses shared native libraries. Consumers must deploy the complete publish output, including the platform-specific shim. Static-link experiments and prerelease package candidates are research evidence, not part of this stable support contract.

## Verification

- Full solution build and Native AOT contract gate: passed
- Device-event focused tests: 37 passed
- Documentation quality and architecture checks: passed
- Fresh `osx-arm64` Native AOT publish and ordinary Core discovery-entry-point run: passed
- Fresh GitHub build, coverage, previews, Native AOT, and dependency-submission checks: passed
- Separate cross-vendor review of the final current-parent diff and toolchain conflict resolution: passed with no blocking findings

## Stack

- Base: #592
- Descendant: #587

Abbreviations: AOT means ahead-of-time compilation; CI means continuous integration; PIV means Personal Identity Verification; SDK means software development kit.

### DIFF
--- a/.github/workflows/native-aot.yml
+++ b/.github/workflows/native-aot.yml
@@ -3,8 +3,9 @@ name: Native AOT Publish Gate (2.0)
 # Publish-and-run verification for the macOS Native AOT host.
 #
 # The host references every supported SDK library, so publishing provides link evidence for all
-# ten libraries. Running it with no attached key provides process-startup and zero-device Core
-# discovery evidence. This workflow does not exercise hardware, monitoring, or applet operations.
+# ten libraries. Running it with no attached key provides process-startup evidence and exercises
+# the ordinary Core discovery entry point. It cannot prove that every discovery transport loaded.
+# This workflow does not exercise hardware, monitoring, or applet operations.
 # See docs/NATIVE-AOT.md for the evidence matrix.
 
 on:
@@ -48,7 +49,7 @@ jobs:
           dotnet publish verification/NativeAotVerification/Yubico.YubiKit.NativeAotVerification.csproj
           -c Release -r ${{ matrix.rid }} --self-contained -p:PublishAot=true
 
-      - name: Run published Native AOT binary (zero-device Core discovery)
+      - name: Run published Native AOT binary (Core discovery entry point)
         working-directory: verification/NativeAotVerification/bin/Release/net10.0/${{ matrix.rid }}/publish
         run: |
           BIN="./Yubico.YubiKit.NativeAotVerification"
--- a/.github/workflows/native-aot.yml
+++ b/.github/workflows/native-aot.yml
@@ -1,15 +1,11 @@
 name: Native AOT Publish Gate (2.0)
 
-# Publish-only verification that the in-scope SDK libraries (Core, Management, Piv, Fido2,
-# WebAuthn, Oath, OpenPgp, SecurityDomain, YubiOtp, YubiHsm) remain Native AOT compatible.
+# Publish-and-run verification for the macOS Native AOT host.
 #
-# This does NOT run tests and does NOT touch CLI projects — see docs/NATIVE-AOT.md for the
-# official support contract. verification/NativeAotVerification is a minimal internal publish
-# host (not a shipped package) that references every in-scope SDK library so the ILC (IL
-# Compiler) trimming/AOT analysis has real code to walk during publish.
-#
-# Phase 1 (current): macOS only, matching the macOS-first support rollout. Phase 2 will extend
-# the matrix to win-x64 and linux-x64 once their runtime verification matrix is closed.
+# The host references every supported SDK library, so publishing provides link evidence for all
+# ten libraries. Running it with no attached key provides process-startup and zero-device Core
+# discovery evidence. This workflow does not exercise hardware, monitoring, or applet operations.
+# See docs/NATIVE-AOT.md for the evidence matrix.
 
 on:
   push:
@@ -52,8 +48,12 @@ jobs:
           dotnet publish verification/NativeAotVerification/Yubico.YubiKit.NativeAotVerification.csproj
           -c Release -r ${{ matrix.rid }} --self-contained -p:PublishAot=true
 
-      - name: Run published Native AOT binary (device discovery only; no hardware required)
+      - name: Run published Native AOT binary (zero-device Core discovery)
+        working-directory: verification/NativeAotVerification/bin/Release/net10.0/${{ matrix.rid }}/publish
         run: |
-          BIN="verification/NativeAotVerification/bin/Release/net10.0/${{ matrix.rid }}/publish/Yubico.YubiKit.NativeAotVerification"
+          BIN="./Yubico.YubiKit.NativeAotVerification"
           chmod +x "$BIN"
-          "$BIN"
+          output="$($BIN)"
+          printf '%s\n' "$output"
+          test "$(printf '%s\n' "$output" | grep -c '^  Linked module entry type: ')" -eq 10
+          printf '%s\n' "$output" | grep -Fx 'Found 0 YubiKey(s).'
--- a/.github/workflows/native-aot.yml
+++ b/.github/workflows/native-aot.yml
@@ -1,0 +1,59 @@
+name: Native AOT Publish Gate (2.0)
+
+# Publish-only verification that the in-scope SDK libraries (Core, Management, Piv, Fido2,
+# WebAuthn, Oath, OpenPgp, SecurityDomain, YubiOtp, YubiHsm) remain Native AOT compatible.
+#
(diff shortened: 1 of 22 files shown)


DennisDyallo:
Updated with two follow-ups from the review:

1. **Inline IL3058 tradeoff doc** (\`docs(aot)\`, 8bcab73e) — the blanket \`NoWarn;IL3058\` in \`Directory.Build.targets\` is now explicitly annotated in \`Yubico.YubiKit.Core.csproj\` at both the \`IsYubiKitSdkLibrary\` opt-in and the \`System.Reactive\` PackageReference, so a maintainer adding a future dependency is warned that IL3058 is globally suppressed and must manually confirm AOT compatibility.

2. **Throwaway PR-only import smoke matrix** (\`test(aot)\`, 0450766e) — `.github/workflows/sdk-import-smoke.yml` verifies packaged importability from a *fresh consumer app* (not just the source tree). Three jobs: `pack`, `import-managed` (10-module matrix, managed build+run), `import-aot-macos` (all 10 packages, osx-arm64 AOT publish+run). **This file is throwaway** — it lives only on `yubikit-aot` so it runs solely for this PR, and should be deleted before/right after merge.

Both were validated locally end-to-end against the packed nupkgs (managed build+run and a real osx-arm64 AOT publish, zero ILC warnings) before pushing.

DennisDyallo:
## Windows Native AOT verification

Verified PR head `0450766ebb7d731dfa2805a405c72ea1accdaa2e` in a fresh detached worktree on Windows x64 (`Microsoft Windows 10.0.26200`).

Command:

```powershell
dotnet publish verification/NativeAotVerification/Yubico.YubiKit.NativeAotVerification.csproj -c Release -r win-x64 --self-contained -p:PublishAot=true
```

Result:

- Publish succeeded and generated native code for `win-x64`.
- No ILC or analyzer warnings were emitted.
- The only diagnostic was informational `NETSDK1057`, because the available SDK was `.NET 10.0.100-rc.1.25451.107`.
- Ran the resulting native executable directly.
- All 9 downstream module entry types linked successfully, covering all 10 SDK libraries including Core.
- Enumeration-only `YubiKeyManager.FindAllAsync` found **2 physical YubiKeys**, both represented as `CompositeYubiKey`.
- The process completed normally without crashing.
- No application sessions, resets, configuration changes, credential operations, or other writes were performed.

This closes the Windows `win-x64` Native AOT publish and basic physical-hardware discovery check, with the caveat that it was executed using the installed .NET 10 RC1 SDK rather than a stable .NET 10 SDK.

DennisDyallo:
## Final Native AOT verification conclusions

I ran the deferred Linux verification from this PR's exact head (`0450766e`) on `linux-x64` with .NET 10.0.100.

### Linux x64: publish and runtime verified

- The permanent `verification/NativeAotVerification` host published successfully with `PublishAot=true`; no AOT/trimming warnings were emitted.
- The output is a real stripped x86-64 ELF Native AOT executable.
- The executable ran successfully and linked the entry types for all 10 in-scope SDK modules.
- Core's Linux discovery/native-loader path initialized against real hardware and found 2 physical YubiKeys.
- The process ended with `Native AOT verification host completed without crashing.` and exit code 0.
- I also packed all 10 SDK packages and repeated the PR-only package-consumer AOT smoke on `linux-x64`. The fresh consumer resolved all packages from the local NuGet feed, Native AOT-published, linked all 10 entry types, printed `aot-import-smoke ok`, and exited 0.

Commands used:

```bash
dotnet publish verification/NativeAotVerification/Yubico.YubiKit.NativeAotVerification.csproj \
  -c Release -r linux-x64 --self-contained -p:PublishAot=true

./verification/NativeAotVerification/bin/Release/net10.0/linux-x64/publish/Yubico.YubiKit.NativeAotVerification
```

### macOS arm64: publish and broader hardware runtime verified

The committed readiness evidence remains valid: a real `osx-arm64` Native AOT publish produced a Mach-O arm64 executable and ran against 3 physical YubiKeys. It exercised Core discovery/native loading, Management/`DeviceInfo`, PIV over the APDU pipeline, NativeShims PC/SC, and the separate OpenSSL/ARKG probe. All 10 modules are link-covered; the remaining application modules called out in the readiness document remain link-only.

### Windows x64: publish and runtime verified

The preceding Windows verification on this same PR head confirms that the `win-x64` Native AOT publish succeeded without ILC or analyzer warnings, and the resulting native executable ran successfully on Windows 10.0.26200. It linked all 10 SDK libraries, initialized Core discovery, found 2 physical YubiKeys represented as `CompositeYubiKey`, and exited normally. The recorded caveat is that this run used .NET 10.0.100 RC1 rather than the stable .NET 10 SDK; `NETSDK1057` was informational.

### Conclusion

The evidence now supports **Native AOT publish/runtime verification on all three primary desktop RIDs: macOS arm64, Linux x64, and Windows x64**. Runtime depth differs: macOS includes Management and PIV protocol operations; Linux and Windows confirm their platform-specific native-loader/Core discovery paths against 2 physical devices each. Linux additionally has packaged-consumer Native AOT linking for all 10 modules. The documented HID and remaining application-session runtime gaps still apply, so this is not a claim that every protocol path has been exercised under AOT.

Recommended support wording: **Native AOT publish and basic physical-hardware runtime verified on macOS arm64, Linux x64, and Windows x64; broader protocol-runtime coverage remains tracked separately.**

DennisDyallo:
@claude we are now good to remove the test ci build from commit 0450766ebb7d731dfa2805a405c72ea1accdaa2e

DennisDyallo:
OK

PR #587, opened 2026-08-26
test(aot): validate internal static NativeShims candidate

DennisDyallo:
Stacked on #578. This pull request validates an internal static-link NativeShims candidate without changing the released consumer contract.

## Stable consumer contract

Public and local builds continue to resolve `Yubico.NativeShims` 1.16.1. Normal build, pack, attest, and publish paths are unchanged from #578, so SDK packages from those paths depend on released version 1.16.1. Native AOT consumers deploy its shared native shim beside the executable as documented in `docs/NATIVE-AOT.md`.

## Candidate validation

- `Directory.Packages.props` keeps 1.16.1 as the default and permits an explicit exact-version override for validation.
- A separate job restores internal prerelease `1.17.4-prerelease.20260825.3` from the authenticated source and runs only by manual dispatch or on this candidate branch.
- The candidate job publishes `osx-arm64`, rejects shared NativeShims sidecars and dynamic NativeShims/OpenSSL dependencies, executes direct OpenSSL and PC/SC calls, and runs the ordinary verification host.
- Exact candidate versions, feed details, package sizes, and historical run links remain in workflow or research evidence. Stable support and event documentation remain inherited from #578.

## Verification

- Stable build and Native AOT contract gate: passed
- Device-event focused tests: 37 passed
- Fast runtime resilience: 74 passed
- Documentation quality, architecture checks, and workflow parsing: passed
- Candidate `osx-arm64` Native AOT job resolved exact `.3`, linked all ten SDK libraries, emitted no shared NativeShims sidecar or dynamic NativeShims/OpenSSL dependency, and passed direct OpenSSL and PC/SC probes
- Fresh GitHub build, coverage, previews, stable Native AOT, candidate Native AOT, and dependency-submission checks: passed
- Separate cross-vendor review of the final current-parent diff and workflow conflict resolutions: passed with no blocking findings

## Package-feed resolution

Historical run 32943193053 published fixed-version `2.0.0-alpha.2` GitHub packages while the candidate override was active. Those versions were retained. The official anonymous alpha feed was superseded with `2.0.0-alpha.30` by workflow run 33859063166; all ten packages were built and attested from current `yubikit`, and the live Core nuspec depends on released NativeShims 1.16.1.

Abbreviations: AOT means ahead-of-time compilation; PC/SC means Personal Computer/Smart Card; SDK means software development kit.

PR #592, opened 2026-08-26
refactor(core): remove System.Reactive from device events

DennisDyallo:
Removes the shipping SDK's `System.Reactive` dependency while preserving `YubiKeyManager.DeviceChanges` as `IObservable<DeviceEvent>`. Adds `YubiKeyManager.WatchAsync(CancellationToken)` as a dependency-free async-enumerable surface.

## Delivery contracts

- `DeviceEventBroadcaster` publishes synchronously in subscription order. An `OnNext` exception propagates and stops delivery to later observers.
- Completion exceptions are isolated per observer so one subscriber cannot prevent other terminal notifications or cleanup.
- Producers own notification serialization; the broadcaster never holds a lock across arbitrary subscriber code.
- Late subscribers receive `OnCompleted` immediately, and unsubscription uses reference identity.
- Each `WatchAsync` enumeration subscribes lazily and owns an independent bounded buffer.
- Buffer overflow faults only the affected stream; cancellation throws `OperationCanceledException`; source completion ends enumeration normally.
- Consumers that need Reactive Extensions operators can reference `System.Reactive` directly; shipping projects do not reference it and test-only interop coverage verifies composition.

## Documentation

Architecture and usage documentation state current event behavior only. Removed-dependency provenance, rejected alternatives, deadlock/revert history, and deferred-design diary content have been removed.

## Verification

- Device-event focused tests: 37 passed
- Reactive Extensions interop tests: 4 passed
- Full solution build: passed
- Documentation quality and architecture checks: passed
- Fresh GitHub build, coverage, migration-preview, architecture-preview, and dependency-submission checks: passed
- Separate cross-vendor review of the final current-base diff: passed with no blocking findings

## Stack

- Base: `yubikit`
- Native AOT path: #578, then #587
- Event-test path: #609, then #624, then #626

Abbreviations: AOT means ahead-of-time compilation; SDK means software development kit.


DennisDyallo:
OK

PR #596, opened 2026-08-31
fix(fido): send CTAPHID_CANCEL when a ceremony awaiting touch is abandoned

DennisDyallo:
## What this fixes

`CtapHidCancel = 0x11` has been sitting in `CtapConstants` without a single caller anywhere in the SDK. That meant abandoning a ceremony never told the authenticator anything, so it kept holding a user-presence request on a channel nobody was servicing.

Symptoms, seen twice on a 5.8.0 key:

- rapid blinking, unresponsive to touch (not the normal awaiting-touch blink)
- `CTAP1_ERR_CHANNEL_BUSY` (0x06) on later opens
- identity reads stop resolving, so the key enumerates under a PID fallback id instead of its serial
- only a physical re-plug recovers it

Canonical Rust sends the cancel from inside the keep-alive loop (`crates/yubikit/src/platform/hidapi.rs:335`), so this is really just restoring parity.

## How it works

The caller's token is threaded into the keep-alive loop **for signalling only**. Reads keep using the exchange token, so `ExchangeGuard`'s guarantee that a partially framed exchange is never stranded still holds. On abandonment the protocol sends `CTAPHID_CANCEL` once, however many keep-alives follow.

One subtlety worth calling out for review: cancellation is surfaced only *after* the authenticator's answer has been read in full, continuation frames included. Throwing earlier would strand those frames for the next exchange to misread — which is precisely the failure this cancel exists to prevent. If a touch lands at the same moment the caller abandons, that answer is a real multi-packet response rather than the one-byte acknowledgement. There's a dedicated test for that case.

While in here, the two tokens on the private path are now named `exchangeToken` and `callerToken`. The one previously called `cancellationToken` is the one that never cancels, which was a bit of a trap.

## Scope

Core only — `FidoHidProtocol` and its tests. An earlier revision of this branch also threaded an optional keep-alive status observer through Fido2 and `IFidoSession`, but that has no production consumer here and belongs with the WebAuthn credential-prompt work that actually uses it, so it has moved to that branch.

## Known limitation

A hard process kill cannot notify the authenticator and still needs a re-plug. That's inherent to the transport.

## Testing

No-hardware `RuntimeResilience` coverage:

- cancel sent exactly once across one and three keep-alives, and the channel is usable immediately afterwards
- a multi-packet answer is drained in full, so the next exchange reads its own response
- no cancel is sent when the caller doesn't abandon

Verified load-bearing by mutation — removing the cancel send, or moving the throw before the drain, fails exactly the tests that should catch it.

Gates run locally:

- `dotnet toolchain.cs build` — clean
- `dotnet toolchain.cs test` — 12/12 projects
- `dotnet toolchain.cs -- resilience --fast` — 74/74 (required, this touches a Core runtime loop)
- `dotnet toolchain.cs -- test --integration --project Fido2 --smoke` — 30/30 on serial 103 / 5.8.0, HID FIDO and SmartCard
- `dotnet toolchain.cs -- test --integration --project WebAuthn` — 11/11

Still outstanding: a hardware regression test for the cancel itself (start a ceremony, don't touch, cancel, assert the key is immediately reusable). It's deliberately a no-touch test, so it should be automatable — happy to add it here or as a follow-up, whichever reviewers prefer.

### DIFF
--- a/src/Core/src/Protocols/Fido/Hid/FidoHidProtocol.cs
+++ b/src/Core/src/Protocols/Fido/Hid/FidoHidProtocol.cs
@@ -80,7 +80,8 @@ internal class FidoHidProtocol(
                             _channelId!.Value,
                             command,
                             data,
-                            exchangeToken)
+                            exchangeToken,
+                            cancellationToken)
                         .ConfigureAwait(false);
                 },
                 cancellationToken)
@@ -166,10 +167,11 @@ internal class FidoHidProtocol(
         uint channelId,
         byte command,
         ReadOnlyMemory<byte> data,
-        CancellationToken cancellationToken)
+        CancellationToken exchangeToken,
+        CancellationToken callerToken = default)
     {
-        await SendRequest(channelId, command, data, cancellationToken).ConfigureAwait(false);
-        return await ReceiveResponse(channelId, command, cancellationToken).ConfigureAwait(false);
+        await SendRequest(channelId, command, data, exchangeToken).ConfigureAwait(false);
+        return await ReceiveResponse(channelId, command, exchangeToken, callerToken).ConfigureAwait(false);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


DennisDyallo:
ok

PR #597, opened 2026-08-31
fix(toolchain): repair coverage collection and add source-level CRAP scoring

DennisDyallo:
## What this fixes

`dotnet toolchain.cs coverage` was broken for every unit test project. It used `dotnet test --collect:"XPlat Code Coverage"`, a VSTest data collector that requires `Microsoft.NET.Test.Sdk`. All 12 unit test projects run on Microsoft Testing Platform and deliberately omit that package (the csproj files even say so), so every run aborted:

```
Testhost process exited with error:
  package: 'testhost', version: '18.0.1-release-25523-111'
  path: 'testhost.dll'
Test Run Aborted.
```

The comment in `toolchain.cs` claiming the VSTest compatibility layer covered both platforms was not accurate, and `TOOLCHAIN.md` repeated it. Both are corrected here.

## Why coverlet.console

I tried the MTP-native route first. `Microsoft.Testing.Extensions.CodeCoverage` throws `TypeLoadException` against `xunit.v3` 3.0.0 — its `Microsoft.Testing.Platform.MSBuild` 1.7.3 dependency is not binary compatible with `Microsoft.Testing.Platform` 2.3.3. That is a latent problem in the xunit.v3 dependency graph; adding the coverage extension just activates it.

`coverlet.console` instruments out of process and never loads into the MTP extension graph. It is also the only option that writes a per-method `complexity` attribute into the Cobertura output, which any coverage-risk reporting needs.

`coverlet.runsettings.xml` is now the single definition of coverage policy. The VSTest collector reads it directly; `toolchain.cs` projects the same settings onto `coverlet.console` flags so the two paths cannot drift.

## New: `dotnet toolchain.cs crap`

Ranks methods by CRAP (Change Risk Anti-Patterns):

```
CRAP(m) = cc(m)^2 * (1 - cov(m))^3 + cc(m)
```

Complexity is computed from the syntax tree with Roslyn, not read from coverlet. coverlet's `complexity` is `Math.Max(1, branches.Count)` over IL branch outcomes, so a single `if` counts 2. Measured across this repo the ratio to source-level complexity ranges from **0.5x to 6x** with no stable multiplier, and because it follows codegen an SDK upgrade can shift the numbers with no source change. That makes it unsuitable for a metric held over time.

Two correctness details worth calling out:

- **Coverage is matched by file and line range, never by name.** coverlet records an async body under `<Name>d__N::MoveNext` but keeps the original file and line numbers. Name matching scores every async method 0%; span matching reads `PivMetadataProtocol.GetSlotMetadataAsync` at its true 90.0%.
- **Implemented methods absent from all reports are scored 0%, not dropped.** Silently omitting them would hide the least-tested code in the repo. Only members that emit no code (abstract, interface, extern, auto-property) are excluded.

## Verification

| Check | Result |
|---|---|
| `dotnet toolchain.cs build` | Succeeded, 0 warnings |
| `dotnet toolchain.cs test` | 12/12 passed |
| `dotnet toolchain.cs coverage` | 12/12, ~1m28s |
| `dotnet crap.cs --self-check` | 36/36 golden fixtures |
| `dotnet toolchain.cs docs-qa` | Succeeded |
| `dotnet format --verify-no-changes` | clean |

The self-check pins every complexity rule with inline fixtures and anchors one case to a hand-derived real method, so the rules cannot drift silently.

## Current baseline

427 methods at CRAP >= 8.

| module | >=8 | worst | | module | >=8 | worst |
|---|---:|---:|---|---|---:|---:|
| Core | 157 | 1640 | | Cli.Shared | 16 | 72 |
| Cli.Commands | 87 | 600 | | Oath | 8 | 30 |
| Fido2 | 44 | 928 | | SecurityDomain | 7 | 110 |
| Piv | 44 | 702 | | YubiHsm | 4 | 42 |
| WebAuthn | 33 | 650 | | Management | 3 | 17 |
| OpenPgp | 21 | 90 | | YubiOtp | 3 | 12 |

Bands: 8-15 → 183 · 15-30 → 132 · 30-100 → 85 · >100 → 27

## Scope

v1 reports only. No CI gate, no cognitive-complexity second axis, no baseline ratchet — those are deliberately deferred until the baseline shows whether they are needed. Nothing in this PR changes CI behavior.

One caveat for reviewers: most of the methods above CRAP 100 are flat status-word and error-string lookup tables. They are large but not nested, so they are poor refactoring targets despite topping the list. The 15-30 band is the more useful place to look.

## Dependency note

Adds `Microsoft.CodeAnalysis.CSharp` to `Directory.Packages.props`, used only by `crap.cs`. No shipping project references it.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,20 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "coverlet.console": {
+      "version": "10.0.1",
+      "commands": [
+        "coverlet"
+      ],
+      "rollForward": false
+    },
+    "dotnet-reportgenerator-globaltool": {
+      "version": "5.5.11",
+      "commands": [
+        "reportgenerator"
+      ],
+      "rollForward": false
+    }
+  }
+}
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -44,5 +44,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="9.0.0" />
     <PackageVersion Include="Bullseye" Version="5.0.0" />
     <PackageVersion Include="SimpleExec" Version="12.0.0" />
+    <!-- Used by crap.cs for source-level cyclomatic complexity. Not referenced by any shipping project. -->
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.6.0" />
   </ItemGroup>
 </Project>
--- a/TOOLCHAIN.md
+++ b/TOOLCHAIN.md
@@ -211,9 +211,38 @@ dotnet toolchain.cs -- crap --crap-args "--top 50 --min-crap 15"
 dotnet toolchain.cs -- crap --crap-args "--json artifacts/crap/crap.json"
 ```
 
-Verify the complexity rules against their golden fixtures with `dotnet crap.cs --self-check`.
+### Two complexity axes
 
-v1 reports only. There is no CI gate, no cognitive-complexity second axis, and no baseline ratchet yet; those wait until the baseline shows whether they are needed.
+CRAP is driven by cyclomatic complexity, which answers *how many paths* — not *how risky is this to change*. Those diverge badly on lookup tables, so the report also computes **cognitive complexity** (SonarSource rule S3776).
+
+| method | cyclomatic | cognitive | reading |
+|---|---:|---:|---|
+| `SCardException.GetErrorString` | 69 | 1 | a flat status-word table: large, obvious |
+| `LinuxHidIOReportConnection.ParseReportSizes` | 13 | 24 | a nested parser: small, genuinely hard |
+
+Cognitive complexity applies three rules that produce this split: a `switch` increments **once** regardless of arm count, nesting compounds, and readable shorthand (`??`, `??=`, `?.`) is ignored.
+
+Filter with both to get the actionable set, which avoids maintaining a hand-written ignore list of "big but harmless" methods:
+
+```bash
+dotnet toolchain.cs -- crap --crap-args "--min-crap 8 --min-cognitive 15"
+```
+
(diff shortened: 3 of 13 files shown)


DennisDyallo:
Added cognitive complexity as a second axis in `73fe80c7`, which supersedes the "no ignore list" question raised in review.

**Why:** cyclomatic complexity answers *how many paths*, not *how risky is this to change*. On this codebase the two diverge hard:

| method | cyclomatic | cognitive |
|---|---:|---:|
| `SCardException.GetErrorString` | 69 | **1** |
| `CtapException.GetMessage` | 48 | **1** |
| `LibcHelpers.GetErrnoString` | 40 | **1** |
| `LinuxHidIOReportConnection.ParseReportSizes` | 13 | **24** |
| `FidoHelpers.DisplayAuthenticatorInfo` | 24 | **29** |

The first three are flat status-word tables — large but obvious. The last two are nested parsers — small but genuinely hard. Ranking on CRAP alone puts the tables on top and buries the parsers.

Filtering on both axes takes **427 methods at CRAP >= 8 down to 16** that are also above the cognitive threshold, with no hand-maintained ignore list. This matters because name-based heuristics would have gone the wrong way: `ParseReportSizes` and `ParseHidDescriptorBytes` *look* like lookup tables by name but are the two highest genuine-risk methods in the repo.

Implementation follows the SonarSource white paper (rule S3776). `--min-cognitive` defaults to 15, Sonar's own threshold.

Also documented the cyclomatic rules against the [published SonarQube C# spec](https://docs.sonarsource.com/sonarqube-server/user-guide/code-metrics/metrics-definition) so the numbers are comparable to any SonarQube report. SonarQube counts conditional access (`?.`), so this does too. One deliberate deviation is called out in `TOOLCHAIN.md`: `catch` clauses and `when` guards increment here, because textbook McCabe and the original crap4j both count them and this repo has 186 catch clauses.

Self-check is now 51 fixtures, several lifted straight from the white paper's worked examples so the implementation can be checked against the spec rather than against itself.

PR #598, opened 2026-08-31
feat(credentials): add ICredentialPrompt and adopt it in WebAuthn

DennisDyallo:
Adds `ICredentialPrompt`, an asynchronous context-carrying prompt abstraction in Core, and adopts it in WebAuthn while keeping status streams observation-only.

## Credential prompting

- Returning `null` means the user declined. Cancellation and failures throw.
- A successful prompt result transfers ownership to the SDK, which zeroes and disposes it after use.
- Rejected PINs are never resubmitted; each retry requests a fresh secret and receives updated context.
- Retry-counter lookup is informational and cannot replace the original PIN rejection; caller cancellation still propagates.
- `MinLengthCodePoints` reflects the authenticator minimum, while `MaxLengthBytes` reflects the UTF-8 encoded limit.
- `ICredentialPrompt` is an SDK-to-application callback; `ISecureCredentialReader` remains an application-initiated terminal helper.
- `MaxPromptAttempts` limits SDK prompt calls per operation. Authenticator retry state remains separate.
- Late prompt results returned after cancellation are observed, zeroed, and disposed.

## WebAuthn behavior

- `WebAuthnStatus` values report progress and results but never gather input or carry response callbacks.
- No dedicated in-flight touch notification is exposed; user guidance can only be speculative.
- `UserVerificationRequirement.Discouraged` does not request verification unless an authenticator constraint or requested permission requires it.
- RP ID enforcement uses `RpIdValidator` and requires a caller-supplied Public Suffix List-backed checker.
- Extensions use internal static adapters coordinated by `ExtensionPipeline`; encoded maps and parsed outputs require explicit coverage.

## Verification

- WebAuthn unit tests: 182 passed
- Credential prompt retry and late-owner tests: 12 passed
- Existing console credential reader tests: 10 passed
- Documentation quality: 58 active files passed
- Full solution build: passed
- Package build with XML documentation: 10 packages, zero warnings
- Separate cross-vendor review of the final base-to-head change: passed with no blocking findings

Hardware user-presence ceremonies were not rerun during this documentation remediation.

Abbreviations: PIN means personal identification number; RP ID means relying party identifier; SDK means software development kit; UTF-8 means Unicode Transformation Format, 8-bit; XML means Extensible Markup Language.

### DIFF
--- a/docs/migration/v1-to-v2-gaps.md
+++ b/docs/migration/v1-to-v2-gaps.md
@@ -136,9 +136,9 @@ v2 locations: `src/Fido2/src/**`, `src/WebAuthn/src/**`
   **Severity**: Major (Blocker for U2F-only consumers) | **Confidence**: High
 
 - **Feature/API**: `KeyCollector` delegate for PIN/touch/UV prompts on `Fido2Session`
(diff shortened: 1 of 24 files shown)


DennisDyallo:
OK

PR #604, opened 2026-08-31
fix(core): preserve X25519 private values and cover private-key ASN.1

DennisDyallo:
## Summary

- Accepts any exactly 32-byte X25519 private value and preserves its bytes through Public-Key Cryptography Standards number 8 import and export.
- Removes pre-clamping validation from stored X25519 key material; scalar masking remains the responsibility of the operation that consumes the key.
- Rejects unsupported key types and invalid X25519 or Ed25519 lengths.
- Centralizes private-key container version validation and prevents access to owned private bytes after disposal.
- Replaces self-referential tests with Base Class Library or published Request for Comments vectors, including X25519 and Ed25519 coverage.

## Why

The previous X25519 verifier rejected valid Request for Comments 7748 private values unless their stored bytes were already clamped. Public-Key Cryptography Standards number 8 containers preserve private values; they must not silently rewrite those bytes during construction, import, or export.

## Verification

- Core cryptography tests: 109 passed.
- Focused Personal Identity Verification import tests: 3 passed.
- Full repository build passed locally.
- The Curve25519 and Ed25519 slice passed separate cross-vendor implementation review, architecture-fit review, and scoped code audit.

## Stack

Pull request #607 remains stacked on this branch and owns private-key decoder zeroing and ownership documentation.


DennisDyallo:
OK

PR #605, opened 2026-08-31
test(core): cover the public-key ASN.1 encoder and decoder

DennisDyallo:
Adds independent coverage for `AsnPublicKeyDecoder` and `AsnPublicKeyEncoder` without changing production code.

## Coverage

- Covers P-256, P-384, P-521, RSA 2048, and RSA 3072 public keys.
- Exercises all four public `EncodeToSubjectPublicKeyInfo` overloads.
- Uses Base Class Library-produced `SubjectPublicKeyInfo` bytes to test decoding, and gives encoder output back to the Base Class Library to test parsing, avoiding self-round-trip oracles.
- Uses published RFC 8410 vectors for Ed25519 and X25519 because .NET 10 has no matching Base Class Library key types.
- Covers unsupported elliptic-curve point formats and lengths, unsupported algorithm identifiers, and unsupported curve identifiers.

Two pre-existing exception sites still expose unsubstituted resource placeholders, so those tests assert exception type rather than entrenching message text.

The later stacked #611 changes malformed private-key public points from a silently incomplete result to `CryptographicException`; this sibling pull request does not depend on that behavior.

## Verification

- Public-key encoder and decoder tests passed.
- Separate remediation review removed defensive reviewer commentary while retaining independent oracle provenance.

Abbreviations: ASN.1 means Abstract Syntax Notation One; RFC means Request for Comments; RSA means Rivest-Shamir-Adleman.

### DIFF
--- a/src/Core/tests/Yubico.YubiKit.Core.UnitTests/Cryptography/AsnPublicKeyDecoderTests.cs
+++ b/src/Core/tests/Yubico.YubiKit.Core.UnitTests/Cryptography/AsnPublicKeyDecoderTests.cs
@@ -22,12 +22,10 @@ namespace Yubico.YubiKit.Core.UnitTests.Cryptography;
 /// Tests for <see cref="AsnPublicKeyDecoder"/>.
 /// </summary>
 /// <remarks>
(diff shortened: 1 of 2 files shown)


DennisDyallo:
OK

PR #607, opened 2026-08-31
fix(core): zero the private-key copies orphaned while decoding PKCS#8

DennisDyallo:
## Summary

- Clears decoder-owned RSA, elliptic-curve, X25519, and Ed25519 private-key temporaries on success and failure.
- Clears superseded RSA normalization buffers when leading-zero padding creates replacement arrays.
- Validates RSA key length and elliptic-curve identifiers before allocating key-owned copies.
- Clears exported elliptic-curve private values in `CreateFromEcdh` and all temporaries in `CreateFromValue`.
- Documents borrowed inputs, caller-owned raw returns, key-owned views, disposal, and clearing responsibilities across the private-key surface.

## Ownership

PKCS#8 and parameter inputs remain caller-owned and are never modified or cleared. Raw decoder returns and PKCS#8 export arrays are caller-owned and must be cleared by the caller. Disposable private-key objects own their copied private material and clear it on disposal.

Internal observe-only hooks make otherwise unreachable arrays testable without changing public API. Their callbacks must not retain or mutate the observed buffers.

## Verification

- Private-key zeroing tests: 12 passed.
- Core cryptography tests: 121 passed.
- Core library, unit-test, and integration-test projects: built with zero warnings and zero errors.
- Explicit-include formatting: clean.
- Separate cross-vendor DevTeam security review: passed after identifying and closing additional RSA and elliptic-curve factory leaks.

## Stack

- Base: #604
- Descendants: #610 and #611

### DIFF
--- a/src/Core/src/Cryptography/AsnPrivateKeyDecoder.cs
+++ b/src/Core/src/Cryptography/AsnPrivateKeyDecoder.cs
@@ -32,10 +32,10 @@ internal class AsnPrivateKeyDecoder
     /// ASN.1 DER-encoded private key.
     /// </summary>
     /// <param name="pkcs8EncodedKey">
-    /// The ASN.1 DER-encoded private key.
+    /// The borrowed ASN.1 DER-encoded private key. This method does not modify or clear the input.
     /// </param>
     /// <returns>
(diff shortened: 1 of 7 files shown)


DennisDyallo:
OK

PR #610, opened 2026-08-31
fix(core): close the remaining unzeroed private-key leaks

DennisDyallo:
Stacked on #607. This pull request closes the remaining RSA private-key encoder buffer leaks without duplicating the parent branch’s EC factory and RSA normalization fixes.

## Changes

- Allocates owned positive ASN.1 INTEGER content octets for every RSA parameter before passing them to `AsnWriter`.
- Clears each temporary content buffer after `AsnWriter.WriteInteger` consumes it, including when an internal observation hook throws.
- Clears the complete encoded PKCS#1 private-key buffer after it is copied into the PKCS#8 wrapper.
- Leaves caller-owned `RSAParameters` arrays unchanged.
- Removes inherited remnants of the former X25519 clamping verifier and uses INTEGER content-octet terminology rather than describing untagged buffers as complete DER integers.
- Drops duplicate descendant EC factory and RSA normalization tests because corrected #607 owns those behaviors.

## Verification

- `PrivateKeyZeroingFollowupTests`: 1 passed
- Core cryptography namespace: 122 passed
- Separate cross-vendor Anthropic DevTeam review: passed

Abbreviations: ASN.1 means Abstract Syntax Notation One; DER means Distinguished Encoding Rules; EC means elliptic curve; PKCS means Public-Key Cryptography Standards; RSA means Rivest-Shamir-Adleman.

### DIFF
--- a/src/Core/src/Cryptography/AsnPrivateKeyEncoder.cs
+++ b/src/Core/src/Cryptography/AsnPrivateKeyEncoder.cs
@@ -60,7 +60,13 @@ internal static class AsnPrivateKeyEncoder
     /// </summary>
     /// <param name="parameters">The RSA parameters including private key values.</param>
     /// <returns>A byte array containing the ASN.1 DER encoded private key in PKCS#8 format.</returns>
-    public static byte[] EncodeToPkcs8(RSAParameters parameters)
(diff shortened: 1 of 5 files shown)


DennisDyallo:
OK

PR #611, opened 2026-08-31
fix(core): reject malformed EC public points instead of returning a wrong answer

DennisDyallo:
Stacked on #607. This pull request makes PKCS#8 EC private-key decoding reject malformed optional public points explicitly.

## Changes

- Rejects an empty public-key bit string with `CryptographicException` before indexing the point prefix.
- Rejects point encodings whose first octet is not the uncompressed-point marker `0x04`.
- Rejects point encodings whose length is not exactly `1 + 2 * coordinateSize`.
- Preserves valid behavior when the optional public-key field is absent and when optional parameter fields are present.
- Preserves corrected parent handling of PKCS#8 version 1 as valid but unsupported and all private-buffer zeroing.
- Keeps tests focused on current malformed-input behavior without prior-behavior, sibling-decoder, line-number, or scope narrative.

## Verification

- `AsnPrivateKeyDecoderTests`: 32 passed
- Core cryptography namespace: 121 passed
- Separate cross-vendor Anthropic DevTeam review: passed with no findings

Abbreviations: EC means elliptic curve; PKCS#8 means Public-Key Cryptography Standards number 8.


DennisDyallo:
OK

PR #612, opened 2026-08-31
fix(core): replace nine unsubstituted resource placeholders with real exception messages

DennisDyallo:
Stacked on #611. This pull request replaces malformed exception-message placeholders and corrects Linux native error reporting.

## Changes

- Replaces nine unsubstituted `ExceptionMessages.UnsupportedAlgorithm` literals with messages that identify the rejected key type, COSE curve, algorithm OID, or curve OID without exposing key material.
- Restores `ArgumentException.ParamName` for unsupported `KeyType` values and adds focused regression tests for all three corrected `KeyDefinitions` throw paths.
- Describes OID `1.2.840.10045.2.1` as `id-ecPublicKey`, not as an ECDSA-only identifier.
- Removes the hard-coded Microsoft C runtime errno table and its value-pinning tests.
- Uses `Marshal.GetLastPInvokeError` immediately after Linux HID/udev failures and `Marshal.GetPInvokeErrorMessage` for runtime-native text. Each `PlatformApiException` receives the same captured code represented by its message.
- Marks the retained public errno helper Linux-only and documents its immediate-call requirement.
- Preserves #611’s strict malformed-point behavior and behavior-focused tests.

## Verification

- `LinuxUdevHidEventSourceTests`: 15 passed
- `LinuxHidDeviceListenerFaultInjectionTests`: 14 passed
- `KeyDefinitionsErrorTests`: 3 passed
- Core cryptography namespace: 121 passed
- `dotnet toolchain.cs -- resilience --fast`: 74 passed
- Separate cross-vendor DevTeam review: passed with no findings after one correction round

Abbreviations: COSE means CBOR Object Signing and Encryption; EC means elliptic curve; ECDSA means Elliptic Curve Digital Signature Algorithm; HID means Human Interface Device; OID means object identifier; P/Invoke means Platform Invocation Services.


DennisDyallo:
OK

PR #613, opened 2026-08-31
fix(core): skip HID long items instead of misparsing the rest of the descriptor

DennisDyallo:
Fixes Linux HID report-descriptor parsing when a long item appears before relevant short items.

## The defect

HID long items use `0xFE`, a payload length, a long-item tag, and then the declared payload. The parser previously decoded `0xFE` as a short item, consumed the length and tag as value bytes, and left the cursor inside the payload. A valid keyboard descriptor after that point could be lost and the interface omitted from Linux HID discovery.

## The correction

- Complete long items are skipped as one header-plus-payload unit and parsing continues.
- A truncated long-item header or payload terminates the walk safely without reading past the descriptor.
- Short-item behavior remains unchanged.

## Verification

- The former known-defect vector now returns the descriptor values after the long item.
- Coverage includes zero-length and multi-byte payloads, truncated headers and payloads, and descriptors containing only complete long items.
- Existing short-item vectors retain their expected values.
- Fast resilience checks and a separate cross-vendor remediation review passed.

Abbreviations: HID means Human Interface Device.

### DIFF
--- a/src/Core/src/Transports/Hid/HidReportDescriptorReader.cs
+++ b/src/Core/src/Transports/Hid/HidReportDescriptorReader.cs
@@ -31,14 +31,8 @@ internal readonly record struct HidReportItem(int Type, int Tag, uint Value);
 /// = tag).
 /// </summary>
 /// <remarks>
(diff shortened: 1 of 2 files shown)


DennisDyallo:
OK

PR #614, opened 2026-08-31
fix(cli): stop returning PINs as unzeroable strings and make prompts scriptable

DennisDyallo:
Replaces command-line secret strings with an owning byte buffer that can be cleared and makes prompts usable with redirected input. #615 remains stacked on this branch.

## Credential ownership

- `SecureCredential` implements `IMemoryOwner<byte>` and exposes `Memory` sliced exactly to the secret length.
- Interactive and redirected prompt buffers are rented from `ArrayPool<byte>`, bounded to the requested maximum, zeroed before every return, and returned exactly once on decline, failure, or disposal.
- Atomic disposal prevents concurrent double-return and access after disposal.
- Unicode scalar input, including surrogate pairs and backspace, is encoded correctly as UTF-8; malformed pairs fail and clear buffered bytes.
- `PinPrompt.ConfirmMatches` uses fixed-time comparison, and `PinPrompt.Resolve` centralizes command-line-or-prompt selection.
- `null` means no credential was supplied and never represents an input error.

`SecureCredential` mentions `ICredentialPrompt` as a prose integration example, but this branch has no compile dependency on #598 and can merge independently.

## Scripted input

Redirected reads terminate on line feed or end-of-input and discard carriage returns. This preserves consecutive prompts with CRLF input and accepts a credential exactly at the configured limit without consuming pool slack.

## Verification

- Cli.Shared unit tests: 43 passed
- Full build and all unit suites passed during implementation
- Scoped formatting passed
- Separate cross-vendor review of the final pooled-buffer implementation passed with no blocking findings

The stacked #615 removes the final YubiHSM string-password boundary.

Abbreviations: CLI means command-line interface; CRLF means carriage return and line feed; PIN means personal identification number; SDK means software development kit; UTF-8 means Unicode Transformation Format, 8-bit.

### DIFF
--- a/src/Cli.Commands/src/Fido/FidoCommands.cs
+++ b/src/Cli.Commands/src/Fido/FidoCommands.cs
@@ -5,7 +5,6 @@ using Spectre.Console;
 using Spectre.Console.Cli;
 using System.ComponentModel;
 using System.Security.Cryptography;
-using System.Text;
 using Yubico.YubiKit.Cli.Commands.Infrastructure;
 using Yubico.YubiKit.Cli.Shared.Output;
 using Yubico.YubiKit.Core.Devices;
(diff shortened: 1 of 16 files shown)


DennisDyallo:
OK

PR #615, opened 2026-08-31
security(yubihsm): take credential passwords as bytes instead of string

DennisDyallo:
Replaces YubiHSM Auth credential-password strings with caller-owned UTF-8 byte memory, validates session-key contexts before device I/O, and corrects the `GetChallengeAsync` password firmware boundary.

## Session-key contexts

- Symmetric context is exactly 16 bytes: `hostChallenge[8] || hsmChallenge[8]`.
- Asymmetric context is exactly 130 bytes: `epkOce[65] || epkSd[65]`.
- Invalid lengths throw before label encoding, touch lookup, list activity, or APDU transmission.

On firmware 5.6.0 and later, `GetChallengeAsync` obtains the YubiKey value. For symmetric credentials on older supported firmware, the caller generates a fresh 8-byte host challenge. The connector receives that exact value and returns the HSM challenge or EPK-SD plus its cryptogram. The SDK does not implement that connector exchange.

Firmware 5.7.1 added the optional credential-password field to `GetChallengeAsync`. Earlier firmware receives no password TLV even when the caller supplies one. `FeatureGetChallengeWithPassword` names the corrected capability; the former field remains as an obsolete compatibility alias.

## Sensitive ownership

Inputs are borrowed and remain caller-owned. The SDK clears its own padded, encoded, and context copies. `SessionKeys` owns returned session material and clears it on disposal. README and HsmAuthTool examples use connector-supplied values and compile in the full solution.

## Verification

- YubiHSM unit tests: 81 passed
- Firmware-boundary APDU tests: 3 passed
- Full solution build: passed with zero warnings
- Documentation quality: 59 active files passed
- Separate cross-vendor review of the final base-to-head change: passed with no blocking findings

Hardware integration was not rerun during this remediation. Context and challenge shapes were checked against canonical Rust and Python yubikit implementations.

Abbreviations: APDU means application protocol data unit; EPK-OCE means ephemeral public key of the off-card entity; EPK-SD means ephemeral public key of the security domain; HSM means hardware security module; I/O means input/output; SDK means software development kit; TLV means tag-length-value; UTF-8 means Unicode Transformation Format, 8-bit.


DennisDyallo:
OK

PR #616, opened 2026-08-31
fix(piv): repair PivTool compilation after the device metadata move

DennisDyallo:
> ## Merge order: land this before #617
>
> **#617 is stacked on this branch.** It restores PivTool and OtpTool to the solution so CI actually compiles them, and it cannot merge until this one lands. Its base retargets to `yubikit` automatically once this merges.
>
> Chain: **#616 → #617**
>
> **Please merge this with a merge commit, not a squash.** #617 is based on this branch. A squash lands this work as a new commit that #617's branch does not contain, which leaves #617 showing duplicated changes and needing a rebase before it can go in. A merge commit makes #617 retarget cleanly with no further work.

**PivTool has not compiled since June.**

Commit `13486277` moved `DeviceInfo`, `DeviceCapabilities` and siblings from `Yubico.YubiKit.Management` to `Yubico.YubiKit.Core.Devices` and, as its own message says, migrated the "mandatory compile consumers". PivTool was not one — it had already fallen out of the solution five months earlier — so two of its four affected files kept importing the old namespace and nothing ever told anyone.

### The part that wasn't obvious

Repointing those two imports exposed **twelve further errors the compiler had been hiding**. Roslyn abandons method-body binding once the declaration phase has failed, and the missing `DeviceInfo` was a property *type*, so every method-body error in the project was suppressed behind it. The "3 errors" this started as was an artefact of that suppression.

Those twelve are a name collision, not more stale imports. PivTool's menus call a local static class named `Certificates` from inside namespace `Yubico.YubiKit.Piv.Examples.PivTool.Cli.Menus`. C# walks enclosing namespaces **before** consulting `using` directives, so the lookup finds the SDK's own `Yubico.YubiKit.Piv.Certificates` namespace and stops. A namespace has no accessibility, so it wins even though everything inside it is `internal`.

Call sites are now qualified through `PivExamples`, which binds unambiguously. Same target, same arguments, no behaviour change.

No analyzer suppressions were needed — the project builds with **zero warnings**, and `PivTool.csproj` is untouched.

### Scope

This only makes PivTool compile. The reason nobody noticed is fixed in the stacked PR on top of this one.

### Verification

`build --project PivTool` → 0 errors, 0 warnings. Full build clean. All 12 unit suites green (2,325 tests). `dotnet format --verify-no-changes` clean on the changed files. Cross-vendor correctness review: pass.

### Worth knowing

The Piv SDK exposes namespaces `Authentication`, `Bio`, `Certificates`, `Cryptography`, `DataObjects`, `Keys`, `Metadata`. Any PivTool example class sharing one of those names will shadow-collide the same way. Today only `Certificates` does.

### DIFF
--- a/src/Piv/examples/PivTool/Cli/Commands/CliRunner.cs
+++ b/src/Piv/examples/PivTool/Cli/Commands/CliRunner.cs
@@ -372,7 +372,7 @@ internal static class CliRunner
         if (device is null) return 1;
 
         await using var session = await device.CreatePivSessionAsync(cancellationToken: ct);
-        var result = await Certificates.GetCertificateAsync(session, slot.Value, ct);
+        var result = await PivExamples.Certificates.GetCertificateAsync(session, slot.Value, ct);
 
         if (useJson)
         {
@@ -406,7 +406,7 @@ internal static class CliRunner
         if (device is null) return 1;
 
         await using var session = await device.CreatePivSessionAsync(cancellationToken: ct);
-        var result = await Certificates.ExportCertificatePemAsync(session, slot.Value, ct);
+        var result = await PivExamples.Certificates.ExportCertificatePemAsync(session, slot.Value, ct);
 
         if (!result.Success)
         {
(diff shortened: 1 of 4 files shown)


DennisDyallo:
OK

PR #617, opened 2026-08-31
build: put PivTool and OtpTool back in the solution so CI compiles them

DennisDyallo:
> ## Merge order: #616 → this PR
>
> **Stacked on #616**, which repairs the compilation break this PR stops CI from hiding. It cannot merge until #616 lands; the base retargets to `yubikit` automatically once it does.
>
> Review #616 first — this PR's diff is the `.sln` change only.

#616 fixed a break that had been on `yubikit` since June. This fixes the reason it went unnoticed.

`dotnet toolchain.cs build`, which is what CI runs, builds `Yubico.YubiKit.sln`. **Neither PivTool nor OtpTool was in it**, so neither was ever compiled by CI. The `--project` path uses separate filesystem discovery and does reach them, which is why the break was reachable on demand but invisible by default.

PivTool was in the solution until commit `93c1568a` — a change about async disposal on session interfaces — **overwrote its project entry with ManagementTool's**. That looks like a bad merge rather than a decision.

OtpTool has never been in the solution at all. It compiles today only because it happened to already import the namespace the metadata types moved into; it would have broken the same silent way otherwise.

### Change

Both registered under their existing solution folders (`Piv`, `YubiOtp`), matching how FidoTool and ManagementTool are already wired. 30 insertions, 0 deletions.

### The regression check

Before: a full solution build produced `FidoTool.dll` but **not** `PivTool.dll` or `OtpTool.dll`.
After (verified with `bin/` and `obj/` deleted first): all three exist.

### Verification

Full build clean. All 12 unit suites green (2,325 tests). Cross-vendor correctness review: pass.

### Follow-up, deliberately not in this PR

PivTool carries **32 files of pre-existing formatting drift** — precisely because it was outside the solution, so `dotnet format Yubico.YubiKit.sln` never touched it. Those files are now visible to anyone running `dotnet format`. No CI gate runs it, so this does not break the build. Left alone to keep this diff to the solution file.


DennisDyallo:
OK

PR #618, opened 2026-09-01
ci: report coverage and CRAP deltas on pull requests

DennisDyallo:
Posts a per-module coverage and CRAP table as a sticky PR comment, with a delta column answering the question that actually matters on a PR: did this change make it better or worse.

## What it looks like

| module | methods | CRAP | Δ CRAP | ≥8 | cog>15 | coverage | Δ cov |
|---|---:|---:|---:|---:|---:|---:|---:|
| **Core** | 1117 | 8417 | -442 | 153 | 5 | 60.9% | +2.1pp |
| Piv | 191 | 1723 | . | 44 | 3 | 67.8% | . |
| OpenPgp | 156 | 1011 | +1 | 21 | 0 | 56.7% | . |
| Cli.Commands | 188 | 5322 | . | 87 | 2 | 2.5% | . |
| **TOTAL** | 2529 | 21920 | **-440** | 423 | 14 | | |

**CRAP decreased by 440.**

> [!NOTE]
> CRAP increased in: OpenPgp (+1).

That is real output, generated by comparing a snapshot taken before #603/#606 merged against `yubikit` today — the -442 in Core is those two PRs landing, and the +1 in OpenPgp is `VerifyPwAsync` gaining an extracted helper.

## Design decisions worth reviewing

**Comparison is by module aggregate, not by method.** Methods get renamed and moved between files; matching them individually produces confident nonsense after any refactor. Aggregates stay meaningful. A module present on only one side still appears, with its counterpart treated as zero.

**Rendering restores preserved head coverage rather than re-running it.** The report must run *after* returning to head, because `--baseline`/`--markdown` only exist in this commit's `crap.cs`. It also needs coverage data, and re-running the full suite a second time to reproduce output we already had would roughly double the job. The head artifacts are copied to `$RUNNER_TEMP` before the base checkout clobbers them, and restored before rendering.

**The base result is cached on its SHA**, so pushing repeatedly to a PR doesn't recompute it.

**Comment lookup paginates.** On a long-lived PR the bot's own comment falls past the first page of 100, and without pagination it would post a fresh duplicate on every push instead of updating in place.

**The base checkout restores an explicitly captured SHA** rather than `git checkout -`, because `pull_request` events start from a detached merge ref.

**Forks still get the numbers.** A forked PR's `GITHUB_TOKEN` is read-only and cannot comment, so the comment step is guarded — but the report is written to the job summary unconditionally.

## Increases are reported, never enforced

`--fail-on-crap-increase` exists and defaults **off**, so the gate can be switched on later without another code change.

One caution if you do turn a gate on: **don't gate on the `≥8` count.** A fully covered method's CRAP equals its cyclomatic complexity, so a 40-arm dispatch table sits above any sane threshold permanently no matter how well tested it is. During this work a change removed 3,135 CRAP while moving that count by +1. Total CRAP is the metric that tracks effort.

## Cost

Two coverage runs per PR (~3 min added) on the first push, one on subsequent pushes thanks to the base cache.

`crap.cs --self-check` grows 51 → 60 fixtures, covering module grouping, one-sided modules, unchanged modules, and verdict selection. Verified that comparing a report against itself yields all-`.` deltas and an `unchanged` verdict.

### DIFF
--- a/.github/workflows/coverage-crap-report.yml
+++ b/.github/workflows/coverage-crap-report.yml
@@ -53,13 +53,23 @@ jobs:
           dotnet toolchain.cs coverage
           dotnet toolchain.cs -- crap --crap-args "--json artifacts/crap/head.json"
 
-      - name: Preserve head measurement
-        # Copied outside the worktree before the base checkout below replaces artifacts/
-        # and crap.cs/toolchain.cs with the base commit's versions.
+      - name: Preserve head measurement and measurement tooling
+        # Two separate reasons to copy things out before the base checkout:
+        #
+        # 1. artifacts/ and head.json would be clobbered by the base build.
+        #
+        # 2. The measurement tooling must be held CONSTANT across both sides. The base
+        #    commit is measured with the *head* commit's crap.cs, toolchain.cs and coverage
+        #    settings, so only the source differs. Otherwise a PR that touches crap.cs -
+        #    such as this one - compares two different metrics and the delta is meaningless.
+        #    It also avoids a bootstrap problem: a base predating a fix to the coverage
+        #    tooling cannot measure itself.
         run: |
-          mkdir -p "$RUNNER_TEMP/head-coverage"
+          mkdir -p "$RUNNER_TEMP/head-coverage" "$RUNNER_TEMP/tooling"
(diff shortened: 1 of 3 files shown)


DennisDyallo (commented):
ok

PR #619, opened 2026-09-01
security(core): stop abandoning an unzeroed copy of every encoded TLV

DennisDyallo:
> ## Merge order: independent
>
> Nothing is stacked on this, and it is not stacked on anything. It can merge at any time, in any order relative to the other open PRs in this batch.

`TlvHelper.EncodeList` assembled its output in an `ArrayBufferWriter<byte>`, then returned `writer.WrittenMemory.ToArray()`. The copy handed to the caller is fine — **the writer's own array is the problem.** It holds the complete BER-TLV encoding of every value passed in, nothing zeroes it, and it is dropped for the GC to reclaim whenever it gets around to it. The bytes stay readable on the heap, can be duplicated by a compacting collection, and can surface in a crash dump.

### Why it matters

This is the shared encoder for **21 production call sites**:

| Module | Sites | Payload |
|---|---|---|
| `HsmAuthSession` | 11 | credential passwords, key material |
| `SecurityDomainSession` | 3 | cert store, key identifiers |
| `ScpState.Scp11` | 2 | SCP11 key-agreement data |
| `OpenPgp/Kdf` | 1 | KDF material |
| PIV data objects | 4 | card metadata |

Every module that encodes a secret into a TLV left a copy of it behind.

### The fix is subtraction, not zeroing

`Tlv.TotalLength` is the length of the very array `Tlv.AsSpan()` exposes, so the sum over the input is not an estimate that might need to grow — it is the **exact** output size. That makes the staging buffer unnecessary. TLVs are now written straight into the array that is returned.

There is no second copy to leak, so this costs **one allocation and one full copy less** than before, rather than adding zeroing work.

Adding zeroing would also have been awkward: `ArrayBufferWriter` exposes no way to clear the array it is about to abandon short of `MemoryMarshal.AsMemory(writer.WrittenMemory)`, which is what `PivDataObjectProtocol.cs:158` resorts to. Not creating the buffer avoids needing that.

### Two input contracts changed, both now documented

**Empty array** previously threw, because `ArrayBufferWriter` rejects a zero `initialCapacity`:

```
System.ArgumentException : Value does not fall within the expected range. (Parameter 'initialCapacity')
   at System.Buffers.ArrayBufferWriter`1..ctor(Int32 initialCapacity)
```

It now returns an empty buffer, matching `EncodeDictionary`. I read all 21 call sites: none can pass an empty array, so this is unreachable in-tree.

**Null array** previously threw the *same* exception by the same accident, since `AsSpan()` maps `null` to an empty span. Removing the writer would have silently turned that into an empty APDU data field, so `EncodeList` now has an explicit `ArgumentNullException.ThrowIfNull`. Losing that failure mode was the one real regression risk in this change.

### Tests

Because `TotalLength` is now the sole size input, a theory pins the **header arithmetic** behind it — expected sizes are written as literals rather than read back off the `Tlv`, so it fails if either the length-field encoding or the meaning of `TotalLength` changes. Verified to fail (7/7 cases) against an off-by-one mutant of `Tlv.Encode`.

Encoded output is asserted against **literal bytes** for the single- and multi-element cases. A boundary theory covers offset arithmetic and absence of trailing slack at the short-form/long-form transitions (`0x7F`, `0x80`, `0x100`).

The empty-input test fails on the current code and passes after.

### Honest scoping note

The absence of a leak here is **structural** — there is no longer a second buffer — so it is verified by inspection, not by a runtime assertion. I did not add a test pretending to prove a negative about GC-reachable memory.

Deliberately **not** fixed here:

- **`ChainedResponseReceiver.cs:44`** — nothing to fix. `ApduResponse` copies (`ApduResponse.cs:34`) and the writer is already cleared at `:57`.
- **`SecurityDomainSession`** has five writers, not two. The two carrying real secrets — `:484` (SCP03 static ENC/MAC/DEK) and `:691` (SCP11 private key) — already `Clear()` in a `finally`. `:360` and `:752` do not clear, but carry CA identifiers and serial numbers, which are public on the wire.
- **A narrower related issue this PR does not solve:** `ArrayBufferWriter.Clear()` only zeroes the *current* array. A default-constructed writer allocates 256 bytes on first write and doubles from there, abandoning each previous array unzeroed — `Clear()` provably cannot reach those. This does **not** affect the SecurityDomain key paths: `:484` (SCP03, ~67-115 bytes) and `:691` (SCP11, ~55 bytes for the P-256-only private path) never exceed 256, so they never grow and `Clear()` fully covers them. It does affect `:423`, which concatenates raw DER certificates, and `ChainedResponseReceiver.cs:44`, which by definition accumulates more than one APDU's worth of response data. Certificates are public; chained response plaintext is command-dependent. Worth filing separately, but it is not a key-material leak.

### Verification

Full build clean. All 12 suites green (2,341 tests). `dotnet format --verify-no-changes` clean.

### DIFF
--- a/src/Core/src/Utilities/TlvHelper.cs
+++ b/src/Core/src/Utilities/TlvHelper.cs
@@ -74,13 +74,17 @@ public static class TlvHelper
     /// </summary>
     /// <param name="tlvData">List of Tlvs to encode</param>
     /// <returns>
-    ///     BER-TLV encoded list, or an empty buffer if <paramref name="tlvData" /> contains no elements.
+    ///     A caller-owned buffer containing the BER-TLV encoded list, or an empty buffer if
+    ///     <paramref name="tlvData" /> contains no elements. The caller must securely clear the returned
+    ///     buffer when it contains sensitive data.
     /// </returns>
-    /// <exception cref="ArgumentNullException"><paramref name="tlvData" /> is <see langword="null" />.</exception>
+    /// <exception cref="ArgumentNullException">
+    ///     <paramref name="tlvData" /> is <see langword="null" />, or contains a <see langword="null" /> element.
+    /// </exception>
     /// <remarks>
-    ///     A disposed <see cref="Tlv" /> still reports its original <see cref="Tlv.TotalLength" /> while
-    ///     exposing zeroed bytes, so passing one encodes a correctly sized run of <c>0x00</c> rather than
-    ///     throwing. Encode before disposing, or use <see cref="EncodeAndDisposeList" />.
+    ///     This method does not dispose the input <see cref="Tlv" /> objects. The caller remains responsible
+    ///     for disposing them. Source buffers used to construct the TLVs also remain caller-owned and are not
+    ///     cleared by this method.
     /// </remarks>
     public static Memory<byte> EncodeList(Tlv[] tlvData)
     {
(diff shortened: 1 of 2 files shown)


DennisDyallo:
OK

PR #622, opened 2026-09-01
refactor(core)!: remove the three obsolete public members

DennisDyallo:
**BREAKING CHANGE.** Removes `Curve25519PublicKey.CreateFromPkcs8`, `RSAPublicKey.CreateFromPkcs8`, and the `HidUsagePage` enum.

These were the **only** three `[Obsolete]` members in the entire codebase, and none of them has a single caller anywhere in the repo.

## What went

| removed | why |
|---|---|
| `Curve25519PublicKey.CreateFromPkcs8` | one-line forwarder to `CreateFromSubjectPublicKeyInfo` |
| `RSAPublicKey.CreateFromPkcs8` | same |
| `HidUsagePage` (whole file) | superseded by `HidInterfaceType` + `HidInterfaceClassifier` |

Both `CreateFromPkcs8` methods were misnamed as much as deprecated: each takes a `subjectPublicKeyInfo` parameter and decodes SPKI, **not** PKCS#8. That is presumably why they were deprecated rather than kept. Callers should use `CreateFromSubjectPublicKeyInfo`, which is exactly what these forwarded to.

`HidUsagePage`'s own attribute called it misleading, and its doc comment warned that `Keyboard = 1` is actually Generic Desktop rather than a keyboard — a genuine trap. The file contained nothing but the enum.

## Why this is safe

- Zero references. The two apparent `HidUsagePage` hits are an unrelated `public short HidUsagePage` property on `CmDevice` that merely shares the name.
- **No `#pragma warning disable CS0618` anywhere in the repo**, so nothing was quietly suppressing an obsolete-usage warning to keep calling these.
- No documentation references either.
- `build`, `test` (12/12), `resilience --fast` and `docs-qa` all clean.

## On the break

All three are public members of public types, so this is a real API break. The package is `2.0.0-alpha.2` and removing deprecated pre-release API is in scope — flagging it explicitly rather than letting it pass as a tidy-up. There is no API-compat baseline in the repo to update.

After this the codebase has no `[Obsolete]` members at all.


DennisDyallo:
OK

PR #609, opened 2026-08-31
refactor(core): simplify device event broadcaster

DennisDyallo:
## Summary

- Simplifies the internal device-event broadcaster while preserving synchronous ordered delivery, exception propagation, completion, and subscription identity semantics.
- Keeps integration coverage for the observable `DeviceChanges` surface.
- Adds an operator-driven repeated hot-plug test that verifies removal and arrival event liveness after a fresh checkpoint for each requested action.

## Event evidence

The hot-plug test intentionally does not correlate `DeviceId` values. It requires interactive standard input, asks the operator to acknowledge readiness before each checkpoint, then waits for a removal or arrival emitted after the corresponding prompt. No fixed quiet period or identity-settling claim is used.

The exact operator-driven test was not rerun during this remediation because no coordinated hardware ceremony was requested. It remains categorized as `RequiresUserPresence` and `Slow`.

## Verification

- Device-event focused tests: 33 passed
- Fast runtime resilience: 74 passed
- Fresh GitHub build, coverage, and dependency-submission checks: passed
- Separate cross-vendor review of the final current-parent diff: passed with no blocking findings

## Stack

- Base: #592
- Descendant: #624

Abbreviations: SDK means software development kit.

PR #624, opened 2026-09-01
test(core): consolidate test doubles and wait helpers onto shared infrastructure

DennisDyallo:
Stacked on #609. This test-only pull request consolidates repeated Core unit-test doubles and polling helpers without changing production behavior.

## Changes

- Renames the shared inert `StubYubiKey` to `FakeYubiKey` and replaces equivalent nested test doubles while preserving each call site's device identifier and advertised connections.
- Keeps connection-serving doubles local where tests require real connection dispatch or request recording.
- Adds `AsyncWait.WaitUntilAsync` for waits where timeout is a failed precondition and `AsyncWait.TryWaitUntilAsync` where timeout is itself an asserted outcome.
- Preserves existing deadlines and cancellation behavior. The shared polling interval is 20 ms; this increases two prior 10 ms polling loops without changing their multi-second deadlines or monotonic predicates.
- Leaves stateful stabilization and admission-barrier helpers unconsolidated because their predicates are not side-effect-free polling conditions.

The rebase onto corrected #609 does not modify its operator-driven integration test or reintroduce identity-pairing or settling-delay claims.

## Verification

- Complete Core unit suite: 996 total, 993 passed, 3 hardware/platform skips
- Fresh GitHub build, coverage, and dependency-submission checks: passed
- Separate cross-vendor review of the final current-parent diff: passed with no findings

Abbreviations: SDK means software development kit.

### DIFF
--- a/src/Core/tests/Yubico.YubiKit.Core.UnitTests/Devices/CompositeDeviceMergerTests.cs
+++ b/src/Core/tests/Yubico.YubiKit.Core.UnitTests/Devices/CompositeDeviceMergerTests.cs
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Yubico.YubiKit.Core.Abstractions;
 using Yubico.YubiKit.Core.Devices;
+using Yubico.YubiKit.Core.UnitTests.Infrastructure;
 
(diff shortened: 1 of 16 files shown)


DennisDyallo:
OK

PR #625, opened 2026-09-01
docs(core): give each device-event contract a single canonical home

DennisDyallo:
Second of the deferred follow-ups recorded on #592. Documentation only — zero behaviour change.

## What

The device-event delivery contracts were written out three times: the XML remarks on the two types,
the "Device Event Delivery" section of `src/Core/CLAUDE.md`, and rows 4–5 of the deferred-decisions
table in `docs/architecture/event-driven-device-discovery.md`. Three copies of a concurrency contract
is three things to keep in sync, and the prose copies had started being where people read instead of
the code.

XML remarks become canonical — they sit with the code they constrain and surface in IntelliSense.
`CLAUDE.md` keeps what an agent needs *before* opening the file and drops the restated reasoning.

## Four claims were promoted before anything was deleted

The interesting finding: several claims flagged as "CLAUDE.md-only" were in the `.cs` file already —
but as inline `//` comments, which does not satisfy "canonical", because an inline comment is
invisible until you have already opened the method. Those were moved into real `<remarks>` first:

- per-observer `OnCompleted` isolation → `Complete()`
- subscription identity / `ReferenceEquals` → `Subscribe()`
- non-blocking `TryWrite` → `DeviceEventStream`

The `Array.IndexOf` warning stays inline as well, because it guards a specific refactor at the exact
line where someone would make it.

## Two contracts stay duplicated on purpose

`OnNext` exceptions aborting later observers, and overflow faulting a stream rather than dropping,
both **read like defects**. A rule belongs where the reader already is, so `CLAUDE.md` keeps them as
imperatives ("Do not wrap the publish loop in a `try`/`catch`") while the reasoning stays canonical in
the XML. They also remain in the architecture doc, which is not redundancy — that table records them
as deferred decisions with reopen triggers, which is a different job.

## The architecture doc is deliberately untouched

Trimming its rows would collide with the deferred-decisions table edited in #578, which adds a row 8
and rewords the paragraph directly above it. Its `Reopen when` column is unique content regardless.
This was the reason the item was deferred on #592 in the first place, and it still holds — so this PR
does the two-thirds that is genuinely unblocked.

## Net +3 lines

Not a reduction, and that is the honest outcome. The XML grew because four claims were promoted from
inline comments into API documentation. The win is not line count — it is that each claim now has one
place it is true from.

## Validation

Build 0 warnings (a malformed `<see cref>` breaks the build here, so every cref was verified to
resolve). 899 unit tests unchanged.

Cross-vendor reviewed (`gpt-5.6-sol`): **PASS**, no findings at any severity. The reviewer enumerated
every claim removed from `CLAUDE.md` and confirmed each is now fully stated in the XML, with line
numbers — that was the one real risk here.

### DIFF
--- a/src/Core/CLAUDE.md
+++ b/src/Core/CLAUDE.md
@@ -218,29 +218,22 @@ Two consumer surfaces over one pipeline, both BCL-typed — the SDK has no react
 
 Two internal types back them, split by reason-to-change:
 
-- `DeviceEventBroadcaster` — multicast only: who is subscribed, and notification order.
-  Copy-on-write observer array; `Publish` reads a lock-free snapshot; mutations take a short lock.
-- `DeviceEventStream` — buffering only: one bounded channel (256) per `WatchAsync` consumer.
-  The publisher uses `TryWrite` and never blocks. Overflow **faults that one stream** rather than
-  dropping, because `DeviceEvent` is a delta and a dropped event permanently desynchronises the
-  consumer's device list.
+- `DeviceEventBroadcaster` (`Devices/DeviceEventBroadcaster.cs`) — multicast only: who is
+  subscribed, and in what order they are notified.
+- `DeviceEventStream` (`Devices/DeviceEventStream.cs`) — buffering only: one bounded 256-event
+  channel per `WatchAsync` consumer.
 
-Contracts worth knowing before editing:
+The delivery contracts — inline synchronous publish, exception handling per notification kind,
+subscription identity, ordering under concurrent publish/complete, and overflow policy — are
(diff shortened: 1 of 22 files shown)


DennisDyallo:
Closing as empty after ancestor corrections. The sole commit moved rejected alternatives, Rx provenance, and deadlock/revert history into implementation XML, which is the placement finding S11 requires removing. Corrected #592 already states every current event-delivery contract in DeviceEventBroadcaster, DeviceEventStream, and the Core agent guide, and #624 preserves them. A separate cross-vendor final-diff review confirmed no unique contract remains and recommended closure rather than padding the pull request.

PR #626, opened 2026-09-01
test(core): retry integration scans that hit transient worker saturation

DennisDyallo:
Stacked directly on #624 after #625 closed empty. This pull request adds a finite integration-test retry for the exact PC/SC worker-admission rejection.

## Behavior

- Retries only an exact `InvalidOperationException` whose full message equals `FindPcscDevices.WorkerSaturationMessage` using ordinal comparison.
- Rejects derived exceptions and near-miss messages immediately so unrelated failures are not masked.
- Makes five total attempts with 200, 400, 800, and 1600 ms delays, then throws with the final saturation exception as `InnerException`.
- Describes this as a finite integration-test policy, not as equivalent to monitor recovery, and reports persistent saturation without diagnosing an unproved wedged worker.

## Test infrastructure

- Places the helper in `Tests.Shared` so every hardware integration suite can use the same policy.
- Wraps the shared lazy device initialization path. Exhausted retries fail the run rather than becoming an empty device list that silently skips every `[WithYubiKey]` test.
- Links the xUnit-free helper source into Core unit tests rather than referencing the xUnit v2-based Tests.Shared assembly from the xUnit v3 test project.
- Wraps only integration scans that can reach PC/SC worker admission; HID-only enumeration remains unwrapped.

## Verification

- `TransientScanRetryTests`: 10 passed
- Complete Core unit suite: 1006 total, 1003 passed, 3 hardware/platform skips
- Full solution build: passed
- Separate cross-vendor review of the final current-parent diff: passed with no blocking findings

The exact hardware integration lane was not rerun during this remediation.

Abbreviations: CI means continuous integration; HID means Human Interface Device; PC/SC means Personal Computer/Smart Card; PR means pull request.


DennisDyallo:
OK

PR #628, opened 2026-09-02
refactor(core): flat published-device model — raw connection slots (Stage C)

DennisDyallo:
## Summary

Completes the flat published-device model refactor: discovery's pre-merge candidates are no longer `IYubiKey` implementations, leaving exactly one lease implementation and one published device shape (`YubiKeyDevice`).

Includes the Stage A/B commit (`fa54aeea`, one routing rule + one published device shape) plus Stage C and its review follow-up.

### Stage C (`08ab6841`)
- Delete `PcscYubiKey`, `HidYubiKey`, `IYubiKeyFactory`, `YubiKeyFactory`
- Narrow `IYubiKeyConnectionSlot` (drops the `IYubiKey` base) — kills the second, production-dead single-interface lease implementation
- Raw connection slots over live enumerated PC/SC and HID handles; interface-id strings byte-identical and pinned by unit tests against real instances
- Hotplug hardening found during cross-vendor audit: unconditional identity+metadata cache eviction on transport activity, generation-checked cache writes (a read started pre-hotplug cannot repopulate a cleared cache), and an atomic `TransportEpoch` so superseded single-flight identity reads fail fast instead of queuing behind hung workers

### Follow-up (`e6291254`)
- Split the slot union into `PcscConnectionSlot` / `HidConnectionSlot` (no nullable transport pair, no runtime transport branch)
- Restore the device-lifetime pin as `YubiKeyDevice_OwnsNoDisposableState`
- `device-identity.md`: caveat that `IYubiKey` dictionary/set keying is reference equality, valid only within one manager's live snapshot (identity-based equality is a Stage D decision)

## Verification
- Core unit: 986 passed; full unit suite: all 12 projects green; resilience `--fast`: 74 passed
- Hardware (5 allow-listed keys): `CompositeDiscoveryIntegrationTests` 6/6, `PivSessionContentionTests` 5/5 (non-user-presence)
- Plan cross-vendor audited (2 rounds), implementation through a 3-round engineer/reviewer loop (final verdict: approve)

## Out of scope (Stage D, separate approval)
Identity policy: interface-set equality vs serial-first correlation, one-slot `DeviceId`, public `DeviceInfo` exposure — preconditioned on metadata-read hardware latency measurement and a failure-backoff decision.

## Post-integration verification (2026-09-04)

[...]

PR #629, opened 2026-09-03
feat(core): define device identity and retention contracts

DennisDyallo:
Stacked on #628. This pull request defines public device identity and repository-retention behavior for the flat published-device model.

## What ships

**Object retention.** Within one manager, an attached key with unchanged interface and connection sets and no contradictory known serial is represented by one retained `IYubiKey` object across scans. Republication as a new object occurs on interface-set change, connection-set change, an observed removal and reinsertion, or fresh known serial evidence proving same-slot substitution.

**Public serial number.** `IYubiKey.SerialNumber` exposes the discovery-read hardware serial without opening a session or depending on the Management package. It remains `null` until discovery establishes a serial, may remain `null` permanently, never regresses to `null`, and is retained on the object delivered by a removal event. A republished object inherits no identity from its predecessor.

**Tri-state physical correlation.** `IYubiKey.SameDeviceAs` evaluates reference identity first: the same reference is `Same`, including when its serial is unknown. Distinct references with equal known serials are `Same`, unequal known serials are `Different`, and an unknown serial on either distinct reference produces `Unknown`. This is not collection equality, so no equality comparer ships.

**Bounded retry behavior.** Failed identity and metadata reads are retried on later scans because failures are not cached. Within one scan, an ambiguous interface identity read makes at most three attempts, each with a two-second caller-wait budget; a timeout is not retried, while other failures use 150 ms and 300 ms retry delays. Best-effort metadata instead makes one no-retry pass over the device's transports with a single three-second caller-wait budget shared across that pass. A device that used the identity-read path does not also incur the metadata-read budget in that scan.

**Referential equality and identifiers.** `Equals` and `GetHashCode` remain referential. Single-interface devices retain transport-shaped `pcsc:*` or `hid:*` identifiers. Repository retention requires complete interface and connection set equality, guarded by contradictory known serial evidence.

## Known bound

Same-slot substitution requires fresh evidence. Transport activity globally invalidates discovery identity and metadata caches before the triggered rescan. If no listener activity is delivered and no scan observes the slot empty, cached evidence can make the substitution unknowable; a caller orchestrating the swap must scan while the slot is empty or restart the manager.

## Verification

- `DeviceIdentityContractTests`: 15 passed
- `YubiKeyDeviceRepositoryCompositeTests`: 19 passed
- `FindYubiKeysFaultInjectionTests`: 16 passed
- `DiscoveryIdentityReaderTests`: 1 passed
- `dotnet toolchain.cs -- resilience --fast`: 75 passed
- `dotnet toolchain.cs docs-qa`: 59 active documentation files validated
- `dotnet toolchain.cs docs-architecture`: passed
- `dotnet toolchain.cs -- pack --include-docs`: full build and ten packages passed with zero warnings and errors
- Separate cross-vendor DevTeam review: passed with no findings after one correction round

The operator-assisted hot-plug test was not rerun during this documentation remediation. Its prior hardware result remains recorded in `docs/architecture/device-identity.md`.

Abbreviations: SDK means software development kit; USB means universal serial bus.

## Post-integration verification (2026-09-04)

Cascading rebase onto integrated #628 preserved all eight original identity commits and adds reviewed integration follow-ups for the current event/test contracts, authorized hardware fixtures, and finder-owned discovery evidence scopes. The latter atomically couples each finder's identity cache, metadata cache, and single-flight read scope, preventing stale reads from crossing transport invalidation or manager replacement.

- Final head: `75a2a800c2454bfbaa5d65b38a926f2e93588324`
- Core unit tests: 1,168 passed, 3 expected skips
- Full unit suite: all 12 projects passed
- Discovery single-flight tests: 13 passed
- Finder fault-injection tests: 16 passed
- Fast resilience: 75 passed
- Full solution build and ten-package build: passed
- Documentation quality and architecture gates: passed
- Fresh GitHub checks: passed
- Final cross-vendor review: passed with no findings

Hardware integration and the operator-assisted hot-plug test were intentionally not rerun because other agents are concurrently using the YubiKeys. Existing hardware evidence remains recorded in the identity documentation.


DennisDyallo:
Update: all stage D' preconditions are now closed with a human-operated hardware session (2026-09-03, macOS, keys 103/31683481/125, OMNIKEY 5022 NFC reader).

**Hot-plug protocol (precondition 3)** — the full remove/reinsert ceremony passed, both via a live console probe and via the committed `HotPlugIdentityContractTests` harness:
- clean `Added` → `Removed` → `Added` timeline, no phantom events (the device-identity question-8 anomaly did not reproduce)
- the `Removed` event delivered the exact retained object with its last-known serial
- reinsertion published a new object; serial re-read from hardware after full cache eviction within the same scan
- `SameDeviceAs` correlated pre/post objects as `Same` in both directions

**Latency measurements + retry policy (preconditions 1–2, R4)** — decided: retry-every-scan stands, no backoff ledger.
- cold 3-key scan (2 USB composites + 1 NFC key, all serials): 260–580 ms; warm scans 1–3 ms; post-hotplug eviction re-read ~250 ms, serial present at publication
- the feared lone-NFC hot loop has no exposure on current firmware: non-YubiKey cards (tested with a credit card) are rejected by the ATR allow-list before any budgeted read; a YubiKey with all NFC apps disabled takes its transport silent entirely; Management cannot be disabled while NFC is up, so a published-but-unreadable NFC key is not manufacturable
- the remaining failure path (published USB device, failing identity read) costs at most 3 fail-fast attempts per scan — now pinned by `FindAllAsync_PersistentIdentityFailureAcrossScans_PerScanAttemptsStayConstant_Pin` (no hardware, never self-skips)

One open observation recorded in the docs: a single earlier ceremony missed a human-confirmed removal when the harness awaited only the event stream under the VSTest host, while the identical sequence in a console host detected it. Not yet reproduced with diagnostics; the harness now records which layer observed the removal so a recurrence will localize itself.

DennisDyallo:
Implemented the substitution follow-up in `66bad175`.

The repository now checks fresh known serial evidence before propagating metadata into a retained object. If otherwise-correlated snapshots have unequal known serials, it replaces the cache entry and emits `Removed` for the predecessor followed by `Added` for the successor. `Unknown` still retains without guessing.

I also corrected two details while grounding the report:
- the vulnerable two-hint window is the 200 ms debounce quiet period; the one-second value caps sustained coalescing
- repository retention also applies to initial and forced manager rescans, not only while monitoring

Coverage includes the production-shaped serial-only substitution, an overriding `SameDeviceAs` implementation that cannot suppress repository-owned serial evidence, independent connection changes, and simultaneous connection/serial changes producing exactly one event pair. The old object retains its original serial and metadata.

The identity and discovery docs now describe four republication triggers and explicitly name the residual case: without a listener notification or a scan observing absence, stale same-interface cache evidence cannot reveal the replacement. A forced scan only after reinsertion is insufficient.

Fresh verification: all 12 unit projects passed (Core total 1009; 1006 succeeded and 3 environment-dependent skips), 75/75 fast resilience tests passed, architecture docs validation passed, and final simplification plus code-audit passes reported no findings.

DennisDyallo:
OK

PR #639, opened 2026-09-04
fix(fido2): preserve forward compatibility

DennisDyallo:
## Summary

- tolerate valid but unmodeled COSE key labels, parameter shapes, algorithms, curves, and key types while preserving raw bytes for unsupported key types
- expose complete owned `RawData` on non-secret response owners: `AuthenticatorInfo`, `CredentialMetadata`, `RelyingPartyInfo`, `FingerprintSensorInfo`, and `EnrollmentSampleResult`
- fix `MakeCredentialResponse.Decode(CborReader)` so it captures exactly one CBOR value and no longer always throws
- document an SDK-wide forward-compatibility doctrine and the FIDO2/WebAuthn escape hatches

## Design boundary

Valid device response data that the SDK does not yet model must not block callers. Malformed CBOR, missing required fields, structurally incomplete modeled keys, and security-critical PIN/UV key agreement remain strict. Unknown response fields are not blindly re-emitted in requests.

`RawData` means an owned copy of a complete non-secret response envelope. `RawCbor` remains the name for encoded nested CBOR structures such as attestation statements and unsupported COSE keys, which may be slices of a parent buffer. We do not add another retained whole-envelope copy to responses containing key material such as `largeBlobKey`; callers needing those raw ceremony responses can use `IFidoSession.SendCborRequestAsync`.

GetInfo option values remain strict Booleans because CTAP defines that shape; unknown Boolean option names remain accepted.

## Audit scope

The FIDO2 applet response surface was audited in parallel across session/PIN/config, credentials/credential management/COSE, and extensions/bio/large blobs. Canonical Rust yubikit, python-fido2, and libfido2 were consulted for biometric wire behavior. That comparison found a pre-existing bio-enrollment mismatch (enumeration should decode top-level key 7 as an array of nested template maps with keys 1/2); it is deliberately deferred rather than folded into this forward-compatibility PR.

## Verification

- `dotnet toolchain.cs -- test --project Fido2` (447 passed)
- `dotnet toolchain.cs -- test --project WebAuthn` (182 passed)
- scoped `dotnet format ... --include ... --verify-no-changes`
- `git diff --check`

No integration or user-presence tests were run; these changes are covered by unit tests and did not contend for shared YubiKeys.

## Review

Completed multiple cross-vendor Engineer/Reviewer loops. Review feedback restored strict non-Boolean GetInfo handling, reduced COSE decoding CRAP from 16.01 to 1, narrowed raw response ownership to avoid duplicate retained key material, and strengthened the CBOR-reader overload test with a packed attestation at a non-zero root offset.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,11 @@
 - `Tlv` and `DisposableTlvList` must be disposed. `TlvHelper.EncodeList` does not dispose its inputs; `EncodeAndDisposeList` does.
 - Listener/native retry loops must block, back off, exit, or throttle on every failure path; add no-hardware fault-injection tests for persistent native errors.
 
+## Forward compatibility doctrine
+- Valid device response data that the SDK does not yet model must not block users. Use open value carriers where practical and preserve complete raw response bytes as an escape hatch.
+- Keep malformed, missing required, and security-critical protocol data strict.
+- Prefer response-side raw escape hatches. Do not blindly re-emit unknown response fields in requests; request encoding remains explicitly modeled and validated.
+
 ## Hardware And Integration Tests
 - Integration tests require authorized YubiKey hardware. The shared test infrastructure reads `YubiKeyTests:AllowedSerialNumbers` from `appsettings.json`; an empty/missing allow list can hard-fail with `Environment.Exit(-1)`, and unauthorized devices are filtered out.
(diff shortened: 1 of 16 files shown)


DennisDyallo:
OK

PR #643, opened 2026-09-04
refactor: Simplify device events, cryptography, YubiHSM, and WebAuthn APIs

DennisDyallo:
## What this does

Alpha API cleanup across device events, Native AOT validation, ASN.1 EC handling, YubiHSM/credential naming, and WebAuthn. Driven by an agreed ideal-state artifact with 54 acceptance criteria; all 54 have direct evidence.

Based on `yubikit` @ `c3b7e206`. Built as five parallel work quanta in isolated worktrees, each cross-vendor reviewed, then merged with zero conflicts, followed by a simplification pass, a whole-diff audit, and one owner-requested follow-up.

Breaking changes are intentional and unwrapped — this is alpha, and no compatibility shims were added.

## Changes by area

**Device events** — `WatchAsync` is now the only public device-change stream. Public `DeviceChanges` (`IObservable`) is gone, along with `DeviceEventBroadcaster`, `DeviceEventStream`, and two unused interfaces. One internal `DeviceEventHub` backs delivery, preserving per-watcher bounded buffers, 256-event overflow isolation, lazy subscription, cancellation isolation, and normal completion on disposal. **`System.Reactive` is dropped from the dependency graph entirely.**

**Native AOT** — the hand-maintained `nativeAotSdkProjects` inventory is replaced by filesystem discovery. The contract now also checks the reverse direction (a host-referenced project that stops matching the SDK layout fails), parses `ProjectReference` with `XDocument` instead of a substring match, and forces an entry-type decision for any newly discovered library.

**ASN.1 / EC** — one shared point-shape validator across all four encoders/decoders. Fixes an `IndexOutOfRangeException` on empty public points, adds curve-specific coordinate and private-scalar length validation, rejects non-EC OIDs that previously produced a valid-looking `id-ecPublicKey` encoding, and emits RFC 5915 `[1]` as an explicitly tagged BIT STRING so keys round-trip through the .NET importer for P-256/384/521.

**WebAuthn** — adds `WebAuthnClientOptions` (`CredentialPrompt`, `EnterpriseRpIds`, `MaxPromptAttempts`, default 3). The backend seam is internal, `FidoSessionWebAuthnBackend` is renamed `WebAuthnBackend`, the ceremony status streams and `WebAuthnStatus` hierarchy are removed, and the 1,255-line client is split into partials by concern. A unified interaction model stays deferred; no replacement callback API was introduced.

**Secret-parameter naming** — the `...Utf8` suffix is dropped from public secret parameters across YubiHSM, Oath, OpenPgp, and Fido2. The UTF-8 contract and caller-ownership/clearing rules move into XML docs; several OpenPgp members had no `<param>` docs at all and now do.

## Notable fixes found along the way

- **PIN/UV token was never zeroed.** `ClientPin`'s decrypted token was cloned into `PinUvAuthTokenSession` and the original abandoned to the GC. Ownership now transfers, so exactly one copy exists and it is zeroed deterministically.
- **`ArgumentException.ParamName` named nothing.** Both `ClientPin.ValidatePin` and `HsmAuthSession` reported an internal helper's local name. Now every public entry point reports its own parameter, pinned by tests.
- **Flaky suite root cause.** `YubiKeyManagerStaticTests` were spinning up real PC/SC and HID listeners inside a unit-test project, failing intermittently with `discovery worker capacity is saturated`. Serialised into the existing discovery-admission collection; the suite is now stable across repeated runs.
- **Three sham guards removed** — tests that read as coverage but could not fail.

## Verification

| Gate | Result |
|---|---|
| `dotnet toolchain.cs build` | 0 errors |
| `dotnet toolchain.cs test` | 12/12 projects, **2688 tests, 0 failed** |
| `dotnet toolchain.cs -- resilience --fast` | 74 / 0 |
| `dotnet toolchain.cs native-aot-contract-qa` | 10 SDK libraries + 2 excluded tooling |

Core 1211/0/3 · Fido2 430/0 · WebAuthn 213/0 · Piv 193/0 · YubiOtp 155/0 · Oath 106/0 · OpenPgp 105/0 · Management 91/0 · YubiHsm 85/0 · SecurityDomain 47/0 · Cli.Shared 43/0 · Cli.Commands 9/0

No integration tests were run — hardware was in use by other work. Integration projects compile.

## Known follow-ups (deliberately not in this PR)

- `WebAuthnClient.Registration.cs` allocates `pinUvAuthParam` before `Build*ExtensionsCbor` can throw, so an extension-pipeline failure leaves a per-request HMAC unzeroed. Wants its own change across both ceremony paths.
- `YubiKeyManagerStaticTests` is stable but still non-hermetic; a hermetic fix needs an `internal static` manager-factory seam.
- `verification/NativeAotVerification/Program.cs` is compiled by no build step in the repo; the AOT contract validates it by text matching only.
- `AsnPublicKeyDecoder` throws `NotSupportedException` for an unsupported-but-well-formed curve on decode — reviewed and accepted, but an open taxonomy question.

## Review notes

Reviewers may want to look hardest at `DeviceEventHub` concurrency, the RFC 5915 tagging change, and the PIN token ownership transfer. Each was independently verified cross-vendor, but they are the three places where a subtle mistake would matter most.

---

## Update: merged with `yubikit` (device-identity / flat device model)

`yubikit` advanced 28 commits while this branch was in review, including PR #629 (flat device model + device identity) and PR #641 (agent-doc consolidation). Both are now merged in.

**Merge policy: our event architecture wins, upstream's device-identity behavior is carried in.** Upstream never touched `DeviceEventBroadcaster`, `DeviceEventStream`, or either dead interface — it only *used* them — so the deletions were uncontested. Only ~40 lines of production code genuinely overlapped.

What was carried in from upstream:
- serial-substitution republishing and `DeviceInfo` metadata retention in `UpdateCache`, rewired onto `DeviceEventHub`
- the flat device model (`YubiKeyDevice`, connection slots) intact
- upstream tests written against the deleted observer harness, ported to `WatchAsync`/`DeviceEventWatcher` with every assertion preserved

### Merge verification

Because a silently-dropped upstream behavior would not fail any test, the resolution was verified structurally rather than only behaviorally:

| Check | Result |
|---|---|
| Symbol probes on `YubiKeyDeviceRepository.cs` | 13/13 — ours and theirs both present in the expected counts |
| Test-name contract (frozen *before* resolving) | 77 upstream + 68 ours, **zero real loss** |
| Attribution audit | 16 contested files, every hunk attributed, none unexplained |
| Auto-merge no-loss proof | `patch(ours→merge) == patch(base→theirs)` and vice versa, byte-exact |
| Assertion census on ported files | `DeviceIdentityContractTests` 33→33, `HotPlugIdentityContractTests` 10→10 |
| Mutation proof | dropping `serialContradiction`, `DeviceInfo` retention, or hub delivery each goes **red** — the second *through the ported upstream tests*, proving they are not sham guards |

### Post-merge gate

12/12 projects, **2757 tests, 0 failed** · Core 1261/0/3 · Fido2 449/0 · resilience 75/0 · AOT 10+2 · build 0 errors.

### Still needs a human

`HotPlugIdentityContractTests` is ported and **compiles but was never executed** — it is hardware-gated and the YubiKeys were in use. Please run it against real hardware before merging.


src/Core/src/Credentials/CredentialPromptContext.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// when the protocol reports it; otherwise <c>null</c>.
+    /// </summary>
+    public int? RetriesRemaining { get; init; }
+
+    /// <summary>
+    /// Gets a value indicating whether this request follows a rejected attempt.
+    /// </summary>
+    public bool IsRetry { get; init; }
+
+    /// <summary>
```
DennisDyallo, line 49: ...

DennisDyallo:
Hardware verification complete — `HotPlugIdentityContractTests` now has a real hardware run behind it, not just a compile check.

**Rig:** single authorized USB YubiKey, SN 103, FW 5.8.0.

**Control run first** (no human interaction, to prove the test is not a sham guard): waited the full 120 s and failed correctly with `No Removed event within 120 s`. The test genuinely requires the ceremony.

**Ceremony run — PASSED (28 s):**

```
[00:00.036] BASELINE    ykphysical:pid:0407  serial=103
[00:25.646] EVENT Removed  serialAtEvent=103  ref=00ca8a98
[00:25.756] REMOVED delivered ref=00ca8a98 retainedSerial=103
[00:27.839] EVENT Added    serialAtEvent=103  ref=0256a4f7
[00:28.019] REINSERTED     serial=103         ref=0256a4f7
watcher pump ended with OperationCanceledException
```

The distinct object references (`00ca8a98` → `0256a4f7`) confirm both contracts over real hardware: `Removed` delivers the **retained** published object with its serial intact, and reinsertion publishes a **new** object that still correlates as `DeviceCorrelation.Same`. Events flowed through the merged `DeviceEventHub`.

This also validates the lazy-subscription fix applied during the port — `WatchAsync` does not subscribe until the first `MoveNextAsync()`, so the enumerator is started on the test thread before `StartMonitoring`. Without it the watcher would have missed the initial scan.

One caveat worth recording: an earlier attempt appeared to pass in 3 s immediately after the rig was re-arranged, because discovery was still settling and residual insertion events satisfied both waits. Valid test, invalid ceremony. Anyone re-running this should let the rig settle first and check the timeline timestamps look human-paced.

PR #645, opened 2026-09-06
fix(webauthn): compute the PIN/UV auth parameter last so it cannot be stranded

DennisDyallo:
Follow-up from #643. A fit audit of the WebAuthn ceremony's handling of secret material turned up one real leak and three smaller misfits around it. All four are here; a fifth finding was considered and rejected, noted at the bottom so it does not get re-raised.

## The leak

Both request builders computed the PIN/UV auth HMAC and *then* did work that can throw — extension CBOR building, and the `PubKeyCredParams` mapping inside the object initializer — before returning the request.

Ownership of that tag transfers to the caller only on a successful return, where `ExecuteMakeCredentialAsync` / `ExecuteGetAssertionAsync` zero it in a `finally`. On a throw the request never reached those callers, so the tag stayed live.

It is reachable without malformed input: LargeBlob `Required` during registration, and any LargeBlob input during assertion, are shipped not-yet-implemented paths that throw from inside extension building.

**On severity, honestly:** this is a marginal security issue, not an urgent one. What leaks is an HMAC bound to one `clientDataHash`, not the PIN and not the token. Anyone who can read this process's memory can already read the `pinUvAuthToken` itself, which is live in `PinUvAuthTokenSession` at the same moment and strictly more useful. What justifies the change is consistency — every other piece of token-derived material in this slice is zeroed, and a silent exception is a trap for the next reader — plus the fact that the fix is smaller than the bug.

## The fix

Rather than add a `catch` to clean up after the fact, both builders now compute the tag **last**, after everything that can throw. `PubKeyCredParams` is hoisted into a local for the same reason: it was the final throwing statement. The protocol version is read before the tag, too.

The window is empty rather than handled, so there is no cleanup path to get wrong later.

The one case left is `OutOfMemoryException` while allocating the request. Deliberately not handled — zeroing during OOM is not reliably possible and the process is failing anyway.

## Three smaller things from the same audit

**`SensitiveMemory.Zero` had a release-build sham guard.** It asserted array backing via `Debug.Assert` and then silently skipped non-array-backed memory, so a release build would have left the secret live — the one behaviour a zeroing helper must never have. It now goes through `MemoryMarshal.AsMemory`, which yields a writable span whatever the backing store is. The branch is gone rather than fixed. Slicing is preserved.

**`PinUvAuthTokenSession.Token` was a `ReadOnlySpan`,** which cannot cross an await, so the exclude-list pre-flight took a `ToArray` copy of the decrypted token — a second live plaintext token, contradicting the type's own documented invariant that exactly one exists. `Token` is now `ReadOnlyMemory`; the copy and its manual zeroing are gone. The trade-off (memory can be captured past disposal, a span cannot) is documented on the property; the type is internal and all callers use it within the session's lifetime.

**A comment in `WebAuthnBackend.MakeCredentialAsync` was wrong.** It justified the defensive copy by citing multi-call reuse of one parameter, but that only happens on the GetAssertion path via `ExcludeListPreflight`'s chunk loop — registration builds a fresh request per attempt and sends it once. The copy is still correct, for ownership reasons, and the comment now says that instead.

## Tests

`TestPinUvAuthProtocol` previously returned `new byte[16]` — all zeroes — which would let any "this was zeroed" assertion pass without the production code doing anything. It now fills tags with a sentinel and records them.

Four tests: two regression tests asserting no tag is produced when extension building fails, each paired with a **control** test proving the recording actually works. Without the controls, `Assert.Empty` could pass because the fake is broken rather than because the fix works.

Both regression tests were confirmed failing before the change with `Assert.Empty() Failure: Collection was not empty`.

## Considered and rejected

The audit also flagged that `ClientPin` does not zero `encryptedToken` or the CBOR request holding `pinHashEnc`, tagged as high value. On inspection that is **ciphertext encrypted under `sharedSecret`, which is itself zeroed in the same `finally`** (`ClientPin.cs:311`, `:469`). Ciphertext left behind after its key is destroyed is not recoverable, and it is not in the AGENTS.md zeroing taxonomy, which calls out decrypted plaintext specifically. Adding ritual zeroing there would dilute the signal that makes the real zeroing legible, so it is deliberately not done.

## Verification

- `dotnet toolchain.cs build` — 0 errors, 0 warnings
- `dotnet toolchain.cs test` — 12 projects, **2761** tests, 0 failed (WebAuthn 217, was 213)
- `dotnet toolchain.cs -- resilience --fast` — 75 passed
- `dotnet format whitespace` scoped to the changed files — clean, verified with a positive control that the filter actually matches

No hardware paths touched.

### DIFF
--- a/src/WebAuthn/src/Client/PinUvAuthTokenSession.cs
+++ b/src/WebAuthn/src/Client/PinUvAuthTokenSession.cs
@@ -45,15 +45,6 @@ internal sealed class PinUvAuthTokenSession : IDisposable
     /// <summary>
     /// Gets the token bytes.
     /// </summary>
-    /// <remarks>
-    /// <see cref="ReadOnlyMemory{T}"/> rather than <see cref="ReadOnlySpan{T}"/> so the token can
-    /// cross an await: the exclude-list pre-flight needs it across several asynchronous probes. A
-    /// span forced that caller into a defensive <c>ToArray</c> copy, which contradicted the whole
-    /// reason this type exists - that exactly one live plaintext copy of a decrypted token should
-    /// be reachable. The trade is that a caller could hold the memory past disposal and read a
-    /// zeroed, or later reused, buffer. The type is internal and its callers all use it inside the
-    /// session's lifetime, so the copy was the worse of the two risks.
-    /// </remarks>
     /// <exception cref="ObjectDisposedException">The session has been disposed.</exception>
     public ReadOnlyMemory<byte> Token
     {
--- a/src/WebAuthn/src/Client/PinUvAuthTokenSession.cs
+++ b/src/WebAuthn/src/Client/PinUvAuthTokenSession.cs
@@ -43,10 +43,19 @@ internal sealed class PinUvAuthTokenSession : IDisposable
     public IPinUvAuthProtocol Protocol { get; }
 
     /// <summary>
-    /// Gets the token bytes as a read-only span.
+    /// Gets the token bytes.
     /// </summary>
+    /// <remarks>
+    /// <see cref="ReadOnlyMemory{T}"/> rather than <see cref="ReadOnlySpan{T}"/> so the token can
+    /// cross an await: the exclude-list pre-flight needs it across several asynchronous probes. A
+    /// span forced that caller into a defensive <c>ToArray</c> copy, which contradicted the whole
+    /// reason this type exists - that exactly one live plaintext copy of a decrypted token should
+    /// be reachable. The trade is that a caller could hold the memory past disposal and read a
+    /// zeroed, or later reused, buffer. The type is internal and its callers all use it inside the
+    /// session's lifetime, so the copy was the worse of the two risks.
+    /// </remarks>
     /// <exception cref="ObjectDisposedException">The session has been disposed.</exception>
-    public ReadOnlySpan<byte> Token
+    public ReadOnlyMemory<byte> Token
     {
         get
         {
--- a/src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
+++ b/src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
@@ -245,9 +245,6 @@ public sealed partial class WebAuthnClient
 
         try
         {
-            // The session's own memory is passed straight through: it stays live for the whole
-            // pre-flight, and taking a copy here would mean a second plaintext token to keep track
-            // of and clear.
             var matchedExclude = await Internal.ExcludeListPreflight.FindFirstMatchAsync(
                 _backend,
                 options.Rp.Id,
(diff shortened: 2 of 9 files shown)


src/WebAuthn/src/Client/PinUvAuthTokenSession.cs
```diff
@@ -43,10 +43,19 @@ internal sealed class PinUvAuthTokenSession : IDisposable
     public IPinUvAuthProtocol Protocol { get; }
 
     /// <summary>
-    /// Gets the token bytes as a read-only span.
+    /// Gets the token bytes.
     /// </summary>
+    /// <remarks>
+    /// <see cref="ReadOnlyMemory{T}"/> rather than <see cref="ReadOnlySpan{T}"/> so the token can
+    /// cross an await: the exclude-list pre-flight needs it across several asynchronous probes. A
```
DennisDyallo: ..

src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
```diff
@@ -327,14 +312,33 @@ private BackendMakeCredentialRequest BuildMakeCredentialRequest(
...
+        // before the PIN/UV auth parameter exists.
+        var pubKeyCredParams = options.PubKeyCredParams
+            .Select(alg => new PublicKeyCredentialParameters { Algorithm = (CoseAlgorithmIdentifier)alg.Value })
+            .ToList();
+
+        // Computed last, deliberately. Nothing between here and the return can throw, so the
+        // parameter cannot be stranded: the request owns it from construction, and
+        // ExecuteMakeCredentialAsync's finally is what clears it. Computing it any earlier leaks
+        // the tag whenever extension building or option mapping fails, because on that path the
+        // request never reaches that finally.
```
DennisDyallo: remove this

src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
```diff
@@ -327,14 +312,33 @@ private BackendMakeCredentialRequest BuildMakeCredentialRequest(
             ? matchedExclude is not null ? new[] { matchedExclude } : Array.Empty<PublicKeyCredentialDescriptor>()
             : options.ExcludeCredentials;
 
+        // Hoisted out of the object initializer below so that every step which can throw happens
+        // before the PIN/UV auth parameter exists.
```
DennisDyallo: remove

src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
```diff
@@ -327,14 +312,33 @@ private BackendMakeCredentialRequest BuildMakeCredentialRequest(
...
+        // parameter cannot be stranded: the request owns it from construction, and
+        // ExecuteMakeCredentialAsync's finally is what clears it. Computing it any earlier leaks
+        // the tag whenever extension building or option mapping fails, because on that path the
+        // request never reaches that finally.
+        ReadOnlyMemory<byte>? pinUvAuthParam = null;
+        byte? pinUvAuthProtocol = null;
+
+        if (tokenSession is not null)
+        {
+            // Version is read first so the tag is the very last thing to come into existence.
```
DennisDyallo: remove

src/WebAuthn/src/Util/SensitiveMemory.cs
```diff
@@ -28,10 +27,14 @@ internal static class SensitiveMemory
...
-    /// this cannot throw instead because every call site is a finally block, where throwing
-    /// would swallow the exception already in flight.
+    /// <see cref="MemoryMarshal.AsMemory{T}(ReadOnlyMemory{T})"/> is what makes this total: it
+    /// yields a writable span whatever the memory is backed by, so there is no "could not zero
+    /// this one" branch left to get wrong. The previous form asserted array backing in debug
+    /// builds and silently skipped it in release, which is the single behaviour a zeroing helper
+    /// must never have - a release build would have left the secret live. Throwing instead is not
+    /// an option either, because every call site is a finally block where it would swallow the
+    /// exception already in flight. Writing through a read-only view is sound here: every buffer
+    /// reaching this method is secret material the caller allocated in order to destroy.
```
DennisDyallo: Remove

src/WebAuthn/src/Client/WebAuthnClient.Registration.cs
```diff
@@ -244,16 +243,17 @@ private async Task<MakeCredentialResponse> ExecuteMakeCredentialAsync(
             return (null, tokenSession, false);
         }
 
-        // ToArray() creates a copy; pre-flight needs to pass token across async boundary.
-        var tokenCopy = tokenSession.Token.ToArray();
         try
         {
+            // The session's own memory is passed straight through: it stays live for the whole
+            // pre-flight, and taking a copy here would mean a second plaintext token to keep track
+            // of and clear.
```
DennisDyallo: remove

DennisDyallo:
@claude adress my PR comments, commit push.

PR #657, opened 2026-09-13
docs: harmonise applet READMEs and fix stale API samples

DennisDyallo:
## Summary

All nine applet READMEs now share one structure and tone, and every code snippet was checked against current source. Several of the old READMEs documented APIs that do not exist.

Structure (identical headings and order in every module): Requirements, Installation, Getting started, Common operations, User interaction, Constraints, Security notes, Example (omitted for WebAuthn and SecurityDomain, which have no sample tool), Related. Each Getting started fence carries the complete `using` block and binds `IYubiKey device`; later fences in the file assume both.

### Per module

| Module | Before | After |
|---|---|---|
| YubiOtp | no README | new |
| WebAuthn, Oath, OpenPgp | 14-20 line feature-PR appendices | full onboarding |
| Fido2, Management | 377 / 411 lines with broken samples | 184 / 168 lines, samples fixed |
| Piv, SecurityDomain, YubiHsm | good content, ad hoc shape | reshaped, trimmed |
| root README | | module links, YubiHsm install line, PIV quick start fixed |

### Stale samples corrected

- Fido2: `PubKeyCredParam`, `AuthenticatorOptions`, `info.CtapVersion`, `info.SupportsResidentKeys`, object-initialised `PublicKeyCredentialRpEntity`, and session-level `SetPinAsync` / `EnumerateCredentialsAsync` / `WriteLargeBlobAsync` / `EnableEnterpriseAttestationAsync` -- none exist. Replaced with `PublicKeyCredentialParameters.CreateES256()`, `MakeCredentialOptions.ResidentKey`, `info.Versions`, constructors, and the `ClientPin` / `PinUvAuthProtocolV2` token flow with `WithPinUvAuth` on both option records.
- Management: `DeviceConfig.FlagEject` is private; public is `DeviceFlags.TouchEject` (with its non-CCID precondition).
- WebAuthn: `options:` named argument does not exist; it is `sessionOptions:`.
- Root: PIV quick start assigned `Task<ReadOnlyMemory<byte>>` to `byte[]`, signed an unhashed message, and skipped PIN verification.
- Various: `preferredConnection` parameter wording replaced with `SessionCreationOptions.PreferredConnectionType`.

Contributor material (project trees, wire tables, test traits, DI matrices, internals) was removed from READMEs; it remains in each module's `CLAUDE.md`.

### Review

[...]

PR #658, opened 2026-09-14
build: ship a per-module package readme on nuget.org

DennisDyallo:
Stacked on #657. Review that one first; this PR's diff against it is only the packaging change.

## Problem

All ten packages currently share the root `PACKAGE_README.md`, so every nuget.org page renders the same 29-line alpha disclaimer and nothing about what the package does. Someone searching "Yubico PIV" lands on a page that only tells them not to use it in production.

## Change

Each packable module ships `src/<Module>/PACKAGE_README.md`, packed at the package root as `README.md`. 51-59 lines each: what it is, requirements, install, one read-only snippet, links.

These are **hand-written, not sliced from the module README**. A package page answers "is this the right package"; `src/<Module>/README.md` answers "how do I use it". I prototyped the generated-slice approach first and dropped it: it needs a bespoke markdown munger on the release path, it broke three times in ten minutes on prose edge cases, and it would let a docs PR break `pack`. The content it slices (intro, requirements, install) is also the most stable part of the file, so the drift it prevents is the drift least likely to happen.

## Wiring

`Directory.Build.props` computes `$(MSBuildProjectDirectory)/../PACKAGE_README.md` and packs it as `README.md` when present, falling back to the shared root readme otherwise, so a future packable project without one still packs instead of failing `NU5039`.

No `.nuspec` files were added. SDK-style pack already carries the metadata, and a hand-written nuspec would stop MSBuild properties from flowing.

Verified by inspecting produced packages:

| Check | Result |
|---|---|
| `Yubico.YubiKit.Piv.nupkg` root | one `README.md`, no stray `PACKAGE_README.md` |
| nuspec `<readme>` | `README.md` |
| Fallback (module readme temporarily removed) | packs `PACKAGE_README.md`, nuspec matches |
| `YubiHsm` nuspec `<description>` | corrected text present |

## Guard

`docs-qa` now checks every packable module for: the file's existence, its own package id in the install command, the alpha banner verbatim, and absolute-https links only. The link check skips fenced code and covers inline, reference-style, and autolink forms.

Fault-injected before trusting it — a relative inline link, a reference definition, and an `http://` autolink were each reported, while `handlers[i](arg)` inside a fence was correctly ignored.

## Also

`Yubico.YubiKit.YubiHsm`'s package description advertised general "YubiHSM support ... hardware security module operations". The applet stores credentials for authenticating to a YubiHSM 2 and never contacts one. Description corrected; this is prominent package-page metadata.

## Review notes

Cross-vendor (GPT) review raised six items. Four fixed: the Oath claim that secrets never reach the host (false at provisioning), SecurityDomain's missing 5.7.0 allow-list floor, the link check's fenced-code false positives and missed link forms, and the YubiHsm description.

Two declined, with reasoning:

- **Core requirements should list YubiKey NEO**, since `ProductAtrs` recognises it. The repository README states the support surface as YubiKey 4, YubiKey 5, and Security Key series; a package page should not widen a support claim beyond it. Core's bullet now mirrors the root README exactly. If NEO is in fact supported, both documents should change together.
- **Add unit tests for the new validator.** No validator in `toolchain.cs` has tests, and the build script is not covered by any test project, so this would hold the new check to a standard nothing around it meets. Fault injection is the verification, as above. Worth a separate discussion if the build script should get a test project at all.

## Follow-ups

- Documentation deep links point at `blob/yubikit/...`. They resolve today and will need retargeting when that branch merges.
- The alpha banner is duplicated in eleven files by design, so that the `docs-qa` const is the single place to find them all when the SDK exits alpha.


DennisDyallo:
Pushed a fix for a link defect caught in review of the diff.

Every package readme, and the nuget.org project link, pointed at the bare repository URL. The repository default branch is `develop`, which is v1 — so a reader arriving from a v2 alpha package page landed on the v1 SDK, with nothing appearing broken. Silently wrong is worse than a dead link.

- Repository links in all ten module readmes and the root fallback now use `/tree/yubikit`.
- `PackageProjectUrl` now uses `/tree/yubikit`. `RepositoryUrl` stays bare: it is source-control metadata and Source Link pairs it with the exact commit.
- `docs-qa` now rejects an unqualified `github.com/Yubico/Yubico.NET.SDK` link in a package readme, requiring `/tree/<branch>` or `/blob/<branch>/...`.

Verified by reverting one link to the bare form and confirming the check fails with a specific message, then confirming a clean run after restoring it. Also unpacked `Yubico.YubiKit.Oath.nupkg` and confirmed `<projectUrl>` and both readme links are branch-qualified.

The existing follow-up still stands and is now broader: these links, `PackageProjectUrl`, and the module deep links all need retargeting when `yubikit` becomes the default branch.

DennisDyallo:
ok

PR #660, opened 2026-09-14
docs: harmonise root and Core READMEs, drop agent-file references

DennisDyallo:
Stacked on #658 (which is on #657). Review those first; this PR's diff against #658 is three concerns:

## Root README

The install block listed all ten packages and explained `--prerelease` as if it were our policy. It is now one feed line plus one package, with the module table as the list. Sentence-case headings, the shared alpha banner, and the Getting started snippet now match the install example (both Management). 166 -> 93 lines.

## Core README

Core was excluded from #657 as a non-applet. Same structure now, adapted: no User interaction or Example sections, but the same headings, sentence case, no emoji, and every snippet checked against source. 408 -> 150 lines.

The TLV usage examples moved verbatim to `docs/usage/tlv-processing.md`, which `src/Core/CLAUDE.md` already treated as canonical; its anchor was dangling and now points there. Concurrency internals, platform interop, and the key-classes table stay in CLAUDE.md and `docs/architecture`.

## No CLAUDE.md links in human-facing docs

Every module README ended its Related list with a link to `CLAUDE.md`. Those files are instructions for coding agents, not documentation. All ten now point at `docs/DEV-GUIDE.md`. `docs-qa` rejects a `CLAUDE.md` or `AGENTS.md` reference in any README, PACKAGE_README, or `docs/usage` page; verified by injecting one and watching it fail with a specific message.

This fixes references introduced in #657 and this layer. Fixing at the top of the stack rather than rewriting the reviewed layers.

## Review notes

Cross-vendor (GPT) review of the root/Core rewrite raised seven items, all addressed: install example vs snippet mismatch; missing `using Yubico.YubiKit.Piv` for the application-session example; `Configure` before `SelectAsync` in the raw APDU example per `docs/architecture/raw-access-tiers.md`; the one-connection-per-key claim now states it is process-local and scoped to discovery-grouped interfaces (`DeviceConnectionRegistry.cs:21-49`); TLV examples relocated rather than lost; `firmwareVersion` in the raw example explained; alpha banner heading kept as-is since it is shared verbatim across all package readmes and enforced by `docs-qa`.

## Verification

```
```


DennisDyallo (commented):
ok

DennisDyallo (commented):
ok

PR #662, opened 2026-09-14
docs: follow-ups from the README harmonisation stack

DennisDyallo:
Stacked on #660 (on #658, on #657). Four independent concerns, one commit each, all raised as loose ends while reviewing the earlier layers.

## Sentence-case alpha banner

`ALPHA - NOT FOR PRODUCTION` was the one Title Case holdout. Changed in the `docs-qa` constant and all twelve copies together; the guard still rejects divergence (verified by altering one copy). The root README also regains its `scripts/alpha/RELEASE_NOTES.md` link, as one clause. That clause rode along in this commit because it touched the same file; it is a two-line change.

## Stale API claims in contributor docs

- `src/WebAuthn/CLAUDE.md` used a named argument `options:` that does not exist on `CreateWebAuthnClientAsync` (the parameters are `clientOptions` and `sessionOptions`, `IYubiKeyExtensions.cs:57`). Same bug the README had.
- `src/Fido2/CLAUDE.md` showed `protocol.InitializeAsync(session)`, which does not exist on `IPinUvAuthProtocol`, and claimed `Dispose` releases an ECDH key pair; the ephemeral pair lives and dies inside `Encapsulate` (`PinUvAuthHelpers.cs:78`) and the shared secret is caller-owned. Snippet rewritten to the real `ClientPin` token flow.
- `DeviceFlags.cs:43` doc comment named a `FidoU2f` capability; the member is `U2f`. Now `<see cref>` so the compiler catches the next drift. Build clean, 0 warnings.

## OATH firmware floor

README and CLAUDE.md claimed firmware 5.0+. Checked against canonical Python (`yubikit/oath.py`), Rust (`crates/yubikit/src/oath.rs`), and Android (`OathSession.java`): none has a session-open floor for OATH, and neither does this session. Only rename (5.3.1) and SCP03 (5.6.3) are gated; a YubiKey 4 works. The real policy is that the SDK carries no NEO workarounds (the post-`set_key` re-select for firmware < 3.0.0). The docs now say that. No code change; adding an artificial gate would break working YubiKey 4 devices for no reason.

## Single source of truth for the documentation branch

The v2 branch name appeared in ~35 links across the package readmes and in `PackageProjectUrl`. `YubiKitDocsBranch` in `Directory.Build.props` now drives `PackageProjectUrl`, and `docs-qa` fails any package-readme repository link whose branch differs from it, naming file, line, found, and expected. The readmes stay literal markdown because they are packed verbatim into nupkgs; generation at pack time was considered and rejected for the reasons in #658. Flipping the branch is one property change plus the sed that `docs-qa` then dictates. The root `PACKAGE_README.md` fallback is now under the same checks (it was not before).

Fault-injected: `/blob/develop/` in a module readme and `/tree/main` in the root fallback both fail with the expected message; `dotnet pack` still emits `<projectUrl>.../tree/yubikit</projectUrl>`.

## Verification

```
dotnet toolchain.cs build
```

## Not done, by decision

`Yubico.YubiKit.YubiHsm` `PackageTags` (`hsm;hardware-security-module`) left as-is; tags are search terms, not claims.


DennisDyallo:
ok

PR #661, opened 2026-09-14
feat(core): add first-key discovery helpers

DennisDyallo:
## Summary
Adds FindFirstAsync and FindFirstOrDefaultAsync with optional predicates. The former throws when no key matches; the latter returns null. Both reuse existing discovery behavior.

Migrates equivalent first-key selections in tests, sample endpoints, and documentation. Callers needing a forced rescan or full device list retain FindAllAsync.

## Verification
- `dotnet toolchain.cs -- test --project Core --filter "FullyQualifiedName~YubiKeyDeviceManagerTests"` — 25 passed before the additional successful-path test.
- `dotnet toolchain.cs -- test --project Core --filter "FullyQualifiedName~YubiKeyDeviceManagerTests.FindFirst"` — all 6 new behavior tests passed on the final code.
- `dotnet toolchain.cs -- test --integration --project Core --filter "FullyQualifiedName~CoreTests.FindFirstOrDefaultAsync_FindsSmartCardDevice"` — 1 passed with hardware.
- `dotnet toolchain.cs -- build --project Core` — passed.
- `dotnet toolchain.cs -- build --project Tests.TestProject` — passed.
- Scoped formatting verification and `git diff --check` passed. Formatting reported existing IL2026/IL3050 analyzer warnings in the sample application.
- Cross-vendor Craftsman fit and correctness review completed; final correctness verdict passed.

## Behavior
The helpers inherit FindAllAsync caching and cancellation behavior. A warm-cache lookup can complete without observing cancellation. They do not wait for insertion or guarantee stable discovery ordering.

PR #663, opened 2026-09-14
docs: bind the first device with FindFirstAsync in every README

DennisDyallo:
Stacked on #662. One commit, eleven files.

Every Getting started block did `FindAllAsync()` then `devices[0]`, which is exactly the pattern #661 added `FindFirstAsync` to replace: it throws a clear `InvalidOperationException` when no key is present rather than an index error, and it is one line.

- Nine module READMEs: the two lines become `IYubiKey device = await YubiKeyManager.FindFirstAsync();`. No other reference to `devices` remained in any of them. `Core.Abstractions` was already in every directive block.
- Core README keeps `FindAllAsync` visible alongside it, since that is Core's own API to show.
- Root README names all three discovery entry points in one sentence so a reader knows the choices exist.

`FindFirstAsync` verified at `src/Core/src/Devices/YubiKeyManager.cs:355`.

```
```


DennisDyallo (commented):
ok

